PR #93, opened 2015-11-12
CP-1143 Add a "link dependency" task

georgelesica-wf:
## Problem

It would be nice to be able to automatically link in dependencies during development. For example, to use a local copy of `packageA` inside of `packageB`, where `packageB` depends on `packageA`. This is already possible, but it requires editing the `pubspec.yaml`, which is awkward and slow in many cases.

## Solution

```
cd <repo of dependency to link>
ddev link
cd <repo of project to link it into>
ddev link <dependency>
pub get
```

We automagically modify the `pubspec.yaml` to add the dependency as a dependency override. Then when you `pub get` you are guaranteed to see your local version of that library. It handles situations where there is already a `dependency_overrides` section, and the case where one does not exist. To go back to the default version, use `ddev unlink`. Once a dependency has been "registered" (the first `ddev link` with no arguments above) it needn't be done again. Those are stored in a JSON file in the user's home directory. One existing issue is that if the dependency already exists in the dependency overrides for some reason this will result in an invalid YAML file. I'm not sure how big a problem that is.

Originally I wanted to use a YAML library to automatically edit the `pubspec.yaml`, but there were a couple problems. First, there isn't good support in Dart for serializing to YAML, doing so required pulling in a third-party dependency. Second, I couldn't figure out how not to mangle the ordering within the file to produce a clean diff (the "new" `pubspec.yaml` would be in a totally different order than the original, meaning any manual changes would be basically lost.

The first implementation I did directly created a symlink in `packages/`. This way, unlinking the dependency can be done by simply running `pub get`. Of course this is also a weakness of the strategy, since any `pub get` or `pub upgrade` will, possibly prematurely, remove the symlink. The other problem is that Dart has deprecated symlinks.

## Usage

See the demo above.

## Tests

I'd like some advice as to how this can be tested.

## FYI

@evanweible-wf


lib/src/tasks/link_dependency/api.dart
```diff
@@ -0,0 +1,88 @@
...
+  Future<LinkDependencyResult> _run(
+      String packageName, Directory linkTarget) async {
+    if (!hasImmediateDependency(packageName)) {
+      throw new ArgumentError(
+          'Package "$packageName" is not a dependency for this project.');
+    }
+
+    // No link target provided, go one directory up and try the package name.
+    if (linkTarget == null) {
+      linkTarget = new Directory('../${packageName}');
```
evanweible-wf: I'm okay with the assumptions behind this logic, but I think the error message below should be more specific and suggest that the user pass in the path to the project as an argument

evanweible-wf:
@georgelesica-wf this is cool, thanks for PRing it! One concern I have is that the `packages` directory was deprecated in Dart 1.12 and is already opt-out via the `--no-package-symlinks` flag. I _think_ 1.13 may formally introduce that change instead of hiding it behind the flag, but I'm not sure on that timeline.

georgelesica-wf:
Interesting, I remember seeing something about that. I'll read up on it and think about the best way to do this on future versions of Dart.

georgelesica-wf:
@evanweible-wf I've changed the implementation to "parse" the pubspec file. All it really does is look for the `dependency_overrides` section. Seems to work nicely. I'm going to have to ponder whether there are any edge cases worth handling. Worst case scenario is it mangles the file and the user has to revert it, so I don't see it as a huge issue at the moment. I still need to add an "unlink" feature. I'm fencing the dependency overrides I'm adding with comments to assist in unlinking.

travissanderson-wf:
This is an awesome idea!

For what it's worth, the symlinking is still in Dart 1.13 but I think they are still hoping to get rid of it eventually.  I wonder if you could just append a section at the top of the .packages file, depending on how pub consumes the file you might be able to short-circuit to the linked version.  This has the same vulnerability to premature breakage of the linked package though.

Have you reached out to the [dart-lang team](https://github.com/dart-lang/pub/issues) about possibly adding this to pub instead?

georgelesica-wf:
@travissanderson-wf I looked at modifying `.packages` but it appears that pub has a tendency to overwrite it even when you didn't tell it to (they have a disclaimer in the docs that you shouldn't edit it manually). Adding this feature to pub might not be a bad idea. I'm hoping to get it into dart_dev, but then maybe I can work on adding it to pub longer-term.

lib/src/tasks/link_dependency/cli.dart
```diff
@@ -0,0 +1,50 @@
...
+import 'package:dart_dev/src/tasks/cli.dart';
+import 'package:dart_dev/src/tasks/link_dependency/api.dart';
+
+class LinkDependencyCli extends TaskCli {
+  ArgParser argParser = new ArgParser();
+
+  final String command = 'link';
+
+  Future<CliResult> run(ArgResults parsedArgs) async {
+    String packageName = parsedArgs.rest.length > 0 ? parsedArgs.rest[0] : null;
```
evanweible-wf: `parsedArgs.rest.isNotEmpty`

evanweible-wf: `parsedArgs.rest.first`

lib/src/tasks/link_dependency/cli.dart
```diff
@@ -0,0 +1,50 @@
...
+    Directory linkTarget =
+        parsedArgs.rest.length > 1 ? new Directory(parsedArgs.rest[1]) : null;
```
evanweible-wf: `parsedArgs.rest.isNotEmpty`

evanweible-wf: ^ ignore me

lib/src/tasks/link_dependency/cli.dart
evanweible-wf: actually I'm okay with this since you have similar logic below that can't use these convenience getters

lib/src/tasks/unlink_dependency/api.dart
```diff
@@ -0,0 +1,75 @@
...
+  Future<UnlinkDependencyResult> _run() async {
+    var pubspecLines = new File('pubspec.yaml').readAsLinesSync();
+    var pubspecFile = new File('pubspec.yaml').openSync(mode: FileMode.WRITE);
+
+    bool insideLinkBlock = false;
+    pubspecLines.forEach((line) {
+      if (line.startsWith('### begin dart_dev link')) {
+        insideLinkBlock = true;
+      }
+      if (line.startsWith('### end dart_dev link')) {
```
evanweible-wf: Could you make these lines constants since you use them in two different tasks?

evanweible-wf:
@georgelesica-wf could you add this task to the readme? There are a few different spots where it should be listed

lib/src/tasks/unlink_dependency/api.dart
georgelesica-wf: Definitely.

georgelesica-wf:
@evanweible-wf I also added a config option so the user can change where the links registry (JSON file mapping project names to directories) is located.

evanweible-wf:
@georgelesica-wf looks good! Now it just needs a `ddev format` and then a rebase and squash into a single commit.

georgelesica-wf:
@evanweible-wf I always forget you guys squash your commits :-). Done and done. Took the opportunity to add the new commands to the Bash completions as well.

evanweible-wf:
+1 thanks!

@trentgrover-wf @maxwellpeterson-wf @dustinlessard-wf

dustinlessard-wf:
:boom: +1

maxwellpeterson-wf:
+1

evanweible-wf:
@jayudey-wf ready for merge

jayudey-wf:
@georgelesica-wf can you update the terminal output when running the command (currently it just indicates `null`)

as well I noticed when un-linking a local dependency that the dependency override section in the pubspec is not removed

georgelesica-wf:
Sorry, lost track of this. I'll get those changes made ASAP @jayudey-wf!

rmconsole2-wf:
@georgelesica-wf This pull request has merge conflicts, please resolve.

rmconsole2-wf:
@georgelesica-wf This pull request has merge conflicts, please resolve.

dustinlessard-wf:
@georgelesica-wf , is this still desired?

georgelesica-wf:
@dustinlessard-wf I think it is, or at least it would be a nice feature, but in the relatively limited time I've spent thinking about it, I haven't come up with a way to accomplish it that doesn't have significant drawbacks. We've got some tooling work planned for this sprint, so as part of that maybe I can circle back to this and come up with something that isn't terrible.

trentgrover-wf:
closing for now

Issue #103, opened 2015-11-21
Take a look at pkg/completion


kevmoo:
https://github.com/kevmoo/completion.dart

Manually writing completion scripts isn't a lot of fun – this just parties off pkg/args

Might be useful?

evanweible-wf:
@kevmoo hadn't seen that before, thanks for posting it! I'll look into using this

georgelesica-wf:
@kevmoo we're using completion.dart now (hooray!) We'd like to be able to provide a facility for doing something like below in a `.bashrc` or similar, but it doesn't appear to be possible.

```
source "$(pub global run dart_dev bash-completion)"
```

My intention was to just put the generated completion script in `lib/` and then write the path to stdout. However, the `Platform.script` attribute is a URL when the app is run as a global, something like `http://localhost:60270/dart_dev.dart`, so we have no way of getting to `lib/`.

Another option would be to generate the script on the fly and use `eval` instead of `source` in the snippet above, but then we'd need to either install completions.dart globally and run it, or move most of the code in the completions.dart script into `lib/` so the completions could be generated via Dart. I could do a quick PR to make that change, unless there's a workaround I'm not seeing...

kevmoo:
@georgelesica-wf please file an issue on the completion package.

I'm guessing we have a way around this with recently added SDK APIs. Need to look into it.

kevmoo:
@georgelesica-wf Actually, I'd say look at https://pub.dartlang.org/packages/package_config

This isn't anything to do with completion, I think. It's everything to do with getting to the lib dir from a globally activated package, right?

`pkg/resource` would be easier, but this only exposes the contents of the file – not the path.

georgelesica-wf:
@kevmoo Oh cool, actually the content would probably be fine because we could just use the `eval ...` pattern in our documentation instead of `source ...`. I'll give it a try, thanks!

kevmoo:
@georgelesica-wf I am doing work so you can 'pub global activate completion' too.

It has a silly dependency on Platform.script being a file: URI at the moment. I'm fixing that up.

PR #109, opened 2015-12-02
CP-1164 Update readme with zsh instructions for completions.

georgelesica-wf:
## Problem

Completion only works on Bash but some people like Zsh!

## Solution

Add instructions for making the completions work with Zsh.

## FYI

@evanweible-wf - you might want to test this to be sure it works in real life, I tested it on my local Zsh from Homebrew, but since I don't use Zsh normally I don't actually have a `.zshrc` and such and I just punched everything in at a command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,21 @@ fi
 Next time you load a Bash session you'll have basic completions for the `ddev`
 alias described above.
 
+#### Zsh Completion
+
+The Bash completion script will work for Zsh as well, but requires a little
+configuration. The following lines must all be found somewhere (and in this
+order, though they needn't be adjacent to one another) in your `.zshrc` file,
+or a file sourced from it:
+
+```
+autoload -U compinit
+compinit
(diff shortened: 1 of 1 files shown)


evanweible-wf:
@georgelesica-wf I tried this out and it worked for me! We should still look into #103 though, because you have to have a copy of dart_dev installed in order to source the completions script and that could lead to completion mismatches if the version installed in a project differs from the version sourced by your bash/zsh profile.

+1

georgelesica-wf:
@evanweible-wf agreed, that looks nice. Maybe I'll take a look at that on Monday. In the meantime, I just didn't want all your Zsh users to be left out of all the fun ;-)

trentgrover-wf:
+1
@jayudey-wf ready for merge (docs only change)

jayudey-wf:
QA +1 doc change only (evan tested the instructions as well)

Merging into master

Issue #116, opened 2015-12-15
CP-2125 Project refactor


evanweible-wf:
There are several restrictions and limitations of the current project architecture that have manifest themselves in the issue backlog. This issue will serve as a place to centralize the common problems behind these limitations, potential solutions to these problems, and ideas to consider as the time for refactor approaches.

---
#1. No distinct stages in execution of a dart_dev task
## Problems
- The project configuration and hook registration occurs at the same time (in the `tool/dev.dart` file)
- It's difficult/impossible for consumer-defined hooks to be dynamic based on run-time configuration
- Relevant issues: #5 
## Proposed Solution
- Break the execution of a dart_dev task up into three stages:
  - **Configuration:**
    - Loads the consumer's `tool/dev.dart` file to allow the project configuration to be set.
    - Parses the command-line options/flags/args and merges it with the project configuration.
    - Determines the dart_dev task that is going to be run ("test", "format", etc.) 
  - **Hook Registration:** 
    - Provide the consumer with the above configuration info so that it can register their own hooks dynamically (the registered hooks would also be "tasks" - see number 2 below)
    - Note: hook registration would be different than it is now. Instead of registering hooks on each task, there would be a single callback that would be called with the config info and the dart_dev task instance. That task instance would have a lifecycle and that is where the hooks would be registered.
  - **Task Execution:**
    - Call into the dart_dev task ("test", "format", etc.) with the above configuration info so that it can determine its own dependencies (ex: the "test" task may want to depend on a "pub serve" task only if browser tests are being run)
    - At this point, a complete "task tree" of sorts exists and can be executed in the necessary order
#2. Lack of robust task & task lifecycle definitions
## Problems
- Not designed to support indeterminate tasks (ex: "pub serve" task that spins up and then runs in the background)
- There's a disconnect between tasks defined in dart_dev ("test" task, "format" task, etc.) and consumer-defined hooks (before/after), making it difficult to integrate the two.
- Relevant issues: #5, #22 
## Proposed Solution
- Define a base task class that supports the two main types of tasks: determinate and indeterminate.
- Support dependencies (ex: if the test task needs a pub server, the "pub serve" task would be a dependency of the "test task")
  - These dependencies should be able to be added dynamically to support conditional logic.
  - This would require the solution from number 1 above (breaking up the execution into multiple steps)
- Require that consumer-defined hooks extend from or implement one of these two types of tasks.
- Make logging easy for both types of tasks:
[...]

PR #94, opened 2015-11-13
Functional coverage

christopherdaly-wf:
Added functional test coverage
@jayudey-wf

PR #123, opened 2016-01-22
Partially decouple selenium from test/coverage tasks.

evanweible-wf:
@christopherdaly-wf 

I spent some time digging around the code from #94 and my main concern was the tight coupling between the test/coverage tasks and the Selenium server - mainly that those tasks were responsible for spinning up the selenium server for functional tests. I envision a plugin system for dart_dev tasks in the future (post #116), so I was hoping to make the selenium dependency function more like a plugin. The main change in this PR is to create a `SeleniumHelper` class that _acts_ (and for the most part functions) like a plugin would. The exception is that the test/coverage tasks still have to know about it (for now) in order to get the observatory ports from it and to kill the child dartium processes.

There are some other changes as well, but I'll run through the diffs and comment on all of them.


.gitignore
```diff
@@ -10,9 +10,9 @@ pubspec.lock
...
-/test/fixtures/coverage/browser_needs_pub_serve/coverage/
-/test/fixtures/coverage/non_test_file/coverage/
-/test/fixtures/coverage/vm/coverage/
-/test/fixtures/docs/docs/doc/api/
+/test_fixtures/coverage/browser/coverage/
+/test_fixtures/coverage/functional_test/coverage
+/test_fixtures/coverage/browser_needs_pub_serve/coverage/
+/test_fixtures/coverage/non_test_file/coverage/
+/test_fixtures/coverage/vm/coverage/
+/test_fixtures/docs/docs/doc/api/
```
evanweible-wf: I moved `test/fixtures/` to `test_fixtures/` so that there's no longer a conflict between the dart_dev project and the fixtures.

README.md
```diff
@@ -588,4 +554,104 @@ Check out the source of these API methods for additional documentation.
...
+> to version control). You can then pass the path to this local executable to
+> the Selenium helper (see below).
+>
+> ```
+> cd tool/
+> curl http://selenium-release.storage.googleapis.com/2.48/selenium-server-standalone-2.48.2.jar
+> echo '#!/usr/bin/env bash' | tee -a selenium-server
+> echo 'DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"' | tee -a selenium-server
+> echo 'exec java -jar $DIR/selenium-server "$@"' | tee -a selenium-server
+> chmod +x selenium-server
```
evanweible-wf: I still need to verify that these commands work

lib/src/task_process.dart
```diff
@@ -57,43 +57,37 @@ class TaskProcess {
   Stream<String> get stderr => _stderr.stream;
   Stream<String> get stdout => _stdout.stream;
 
-  Future<List<int>> getChildPids(List<int> pids) async {
+  bool kill([ProcessSignal signal = ProcessSignal.SIGTERM]) =>
+      _process.kill(signal);
+
+  Future<bool> killAllChildren(
+      [ProcessSignal signal = ProcessSignal.SIGTERM]) async {
```
evanweible-wf: Previously this method was expecting a process id as an arg, but since this is a method on a `TaskProcess` instance, it makes more sense for it to kill all children of that process. I also added the `signal` param to be consistent with the `kill()` method

lib/src/tasks/coverage/api.dart
```diff
@@ -70,12 +70,11 @@ class CoverageTask extends Task {
...
   static Future<CoverageResult> run(List<String> tests,
-      {bool functional: defaultFunctional,
-      bool html: defaultHtml,
+      {bool html: defaultHtml,
       String output: defaultOutput,
       List<String> reportOn: defaultReportOn}) async {
-    CoverageTask coverage = new CoverageTask._(tests, reportOn,
-        html: html, output: output, functional: functional);
+    CoverageTask coverage =
+        new CoverageTask._(tests, reportOn, html: html, output: output);
```
evanweible-wf: I'm no longer passing the `functional` flag through here (which was used to conditionally spin up the Selenium server). This is because I've passed that responsibility on to the project (they have to use the `SeleniumHelper` class to start and stop the server). The side effect of this is that projects with functional tests will have to run the selenium server on every test run, even if functional tests aren't being run. This will be addressed in the project refactor and I'm fine with it for now because it shouldn't be a ton of overhead.

lib/src/tasks/coverage/api.dart
```diff
@@ -568,11 +550,9 @@ class CoverageTask extends Task {
           break;
         }
       }
-      if (server != null && server.observatoryPort.length > 0) {
-        List<int> result = await server.checkPorts();
-        return result;
-      }
-      return [port];
+      var observatoryPorts = await SeleniumHelper.getActiveObservatoryPorts();
```
evanweible-wf: Since the test/coverage tasks are no longer responsible for starting the selenium server, the API for getting the active observatory ports is now statically available. This is temporary and would be replaced by the plugin system I mentioned before, at which point the test/coverage tasks would be completely decoupled from the selenium dependency.

lib/src/tasks/coverage/exceptions.dart
```diff
@@ -38,10 +38,3 @@ class MissingLcovException implements Exception {
   final String message = missingLcovMessage;
   String toString() => 'MissingLcovException: $message';
 }
-
-class PortBoundException implements Exception {
-  final String message;
-  PortBoundException(String message, String port)
-      : this.message = 'Port $port could not be bound: $message';
-  String toString() => 'PortBoundException: $message';
-}
```
evanweible-wf: I didn't really mean to remove this, but it didn't make sense in this exceptions file since it's not specific to coverage. It can certainly be re-added, would just need to live somewhere else.

lib/src/tasks/coverage/config.dart
```diff
@@ -20,14 +20,10 @@ import 'package:dart_dev/src/tasks/test/config.dart';
...
 const String defaultOutput = 'coverage/';
 const List<String> defaultReportOn = const ['lib/'];
-const String defaultSeleniumCommand = "selenium-server";
-const String defaultSeleniumSuccess = "Selenium Server is up and running";
 
 class CoverageConfig extends TaskConfig {
   bool html = defaultHtml;
-  bool pubServe = defaultPubServe;
-  String seleniumCommand = defaultSeleniumCommand;
-  String seleniumSuccess = defaultSeleniumSuccess;
```
evanweible-wf: Again, decoupling the coverage task from selenium, so I removed the selenium specific stuff

lib/src/tasks/test/api.dart
```diff
@@ -60,11 +52,8 @@ Future<TestTask> test(
     if ((line.contains('All tests passed!') ||
             line.contains('Some tests failed.')) &&
         !outputProcessed.isCompleted) {
-      if (server != null) {
-        await server.closeTests();
-        await server.kill();
-      }
```
evanweible-wf: same stuff here, selenium is no longer responsibility of test task

test/fixtures/coverage/functional_test/lib/index.dart
```diff
@@ -1,11 +0,0 @@
-import 'dart:html';
```
evanweible-wf: The next few files show up as completed removed, but that's only because I changed a few things and renamed the parent `test/fixtures/` directory. Most of this is the same or similar

test/integration/examples_test.dart
```diff
@@ -36,7 +36,7 @@ Future<bool> serveExamplesFor(String projectPath) async {
   process.stdout.listen((line) async {
     if (line.contains(pubServePattern)) {
       served = true;
-      await process.killGroup();
+      await process.killAllChildren();
```
evanweible-wf: I think I had this question on your PR as well - what was this added for?

test/fixtures/coverage/functional_test/test/pubspec.yaml
```diff
@@ -1,7 +0,0 @@
-name: functional_test_changeme
```
evanweible-wf: This file I actually did remove - previously I remember it being necessary to have a separate pubspec.yaml wherever the functional tests were located because it required an older version of the test package - but that no longer seems to be the case. I removed it and it worked, so I think it's fine now?

evanweible-wf: This file I actually did remove - previously I remember it being necessary to have a separate pubspec.yaml wherever the functional tests were located because it required an older version of the test package - but that no longer seems to be the case. I removed it and it worked, so I think it's fine now?

test_fixtures/coverage/functional_test/tool/selenium-server
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+exec java -jar $DIR/selenium-server-standalone.jar "$@"
```
evanweible-wf: I committed a local `selenium-server` executable along with the JAR (below) so that we can get more consistent results on this integration test, and also just to prove that the `SeleniumHelper` class works with a globally installed selenium server or with a local executable.

test/integration/examples_test.dart
christopherdaly-wf: Honestly, I don't remember now.  It doesn't seem necessary though.

christopherdaly-wf:
+1

PR #124, opened 2016-01-27
CP-1329 Create dart test runner generator, dart_dev

jayudey-wf:
## Issue
- Several teams were creating test runners in different ways.  It would be nice to unify this approach across teams.

## Changes
**Source:**
- Added dart_dev task to generate a test runner

**Tests:**
- added

## Areas of Regression
- n/a (failing smithy build from additional tests)

## Testing
- install into a project and ensure that runner is created as expected

## Code Review
@trentgrover-wf
@maxwellpeterson-wf
@evanweible-wf
@dustinlessard-wf


lib/src/tasks/gen_test_runner/config.dart
```diff
@@ -0,0 +1,49 @@
...
+const String defaultFilename = 'generated_runner';
+const bool defaultGenHtml = false;
+const bool defaultReact = true;
+const List<String> defaultScriptTags = const [
+  'packages/react/react_with_addons.js'
+];
+
+enum Environment { vm, browser }
+
+class SingleRunnerConfig {
```
evanweible-wf: The name `SingleRunnerConfig` is a bit confusing to me.. I'll try to throw out some alternatives:

`TestSuite`
`TestSuiteConfig`
`TestRunnerConfig`

test/integration/gen_test_runner_test.dart
```diff
@@ -0,0 +1,97 @@
...
+
+  await process.done;
+
+  return new Runner(await process.exitCode, errors, files);
+}
+
+verifyExistence(String file, bool shouldExist, {bool delete: false}) {
+  expect(FileSystemEntity.isFileSync(file), equals(shouldExist));
+  if (shouldExist) {
+    new File(file).deleteSync();
```
evanweible-wf: #nit Confusing at first that you have a method that verifies the existence of a file and then deletes the file only if it should exist. Maybe rename to `verifyExistenceAndCleanup()` or something

test/integration/gen_test_runner_test.dart
```diff
@@ -0,0 +1,97 @@
...
+  }
+  TaskProcess process =
+      new TaskProcess('pub', args, workingDirectory: projectPath);
+
+  await process.done;
+
+  return new Runner(await process.exitCode, errors, files);
+}
+
+verifyExistence(String file, bool shouldExist, {bool delete: false}) {
```
evanweible-wf: `shouldExist` is an [enigmatic boolean](https://techblog.workiva.com/tech-blog/arrested-development-enigmatic-booleans) here.

evanweible-wf: it might make sense to break this up into two separate efforts - verifying a file exists or doesn't exist, and then removing the file afterwards

test/integration/generated_runner.dart
```diff
@@ -0,0 +1,31 @@
+@TestOn('vm')
+
+/************ GENERATED FILE ************
+
+This file was generated with the command:
+gen-test-runner -d test/integration -e Environment.vm --no-react --no-genHtml
```
evanweible-wf: This would actually be `dart_dev gen-test-runner` or `pub run dart_dev gen-test-runner`, right?

evanweible-wf: or can you run it directly?

lib/src/tasks/gen_test_runner/config.dart
```diff
@@ -0,0 +1,49 @@
...
+  bool react = defaultReact;
+  List<String> scriptTags = defaultScriptTags;
+
+  SingleRunnerConfig(
+      {String this.directory: defaultDirectory,
+      Environment this.env: defaultEnv,
+      String this.filename: defaultFilename,
+      bool this.genHtml: defaultGenHtml,
+      bool this.react: defaultReact,
+      List<String> this.scriptTags: defaultScriptTags});
```
evanweible-wf: "script tags" is a bit misleading - it's not the tags, but the paths to the scripts. Maybe `scriptPaths`?

lib/src/tasks/gen_test_runner/cli.dart
```diff
@@ -0,0 +1,87 @@
...
+        help:
+            'Configuration options should be performed in local dev.dart file');
+
+  final String command = 'gen-test-runner';
+
+  Future<CliResult> run(ArgResults parsedArgs) async {
+    GenResultGroup results = new GenResultGroup();
+
+    for (var i = 0; i < config.genTestRunner.configs.length; i++) {
+      SingleRunnerConfig currentConfig = config.genTestRunner.configs[i];
```
evanweible-wf:
```dart
for (var currentConfig in config.genTestRunner.configs) {
```

test/integration/generated_runner.dart
jayudey-wf: no you can't run it directly as currently implemented, I removed that due to;
- the number of configurations per test runner
- possibly generating multiple test runners per command execution

lib/src/tasks/gen_test_runner/cli.dart
```diff
@@ -0,0 +1,87 @@
...
+          genHtml: currentConfig.genHtml,
+          react: currentConfig.react,
+          scriptTags: currentConfig.scriptTags);
+      await task.done;
+      results.tasks.add(task);
+      results.passing.add(task.successful);
+    }
+
+    if (results.passing.contains(false)) {
+      return new CliResult.fail('Failed to generate test runner');
```
evanweible-wf: Should expand this error message to say _which_ file(s) failed to generate, since it could be one or many from a set.

lib/src/tasks/gen_test_runner/cli.dart
```diff
@@ -0,0 +1,87 @@
...
+        return new CliResult.fail('React tests can not be run on the vm');
+      }
+
+      GenTestRunnerTask task = await genTestRunner(
+          filename: currentConfig.filename,
+          directory: currentConfig.directory,
+          environment: currentConfig.env,
+          genHtml: currentConfig.genHtml,
+          react: currentConfig.react,
+          scriptTags: currentConfig.scriptTags);
```
evanweible-wf: Since you've already established the concept of a single test suite config (currently `SingleRunnerConfig`) and you're forwarding all of the options through to the API method, it's probably fine to change the API to accept an instance of `SingleRunnerConfig` to be more succinct.

lib/src/tasks/gen_test_runner/cli.dart
```diff
@@ -0,0 +1,87 @@
...
+    for (var i = 0; i < config.genTestRunner.configs.length; i++) {
+      SingleRunnerConfig currentConfig = config.genTestRunner.configs[i];
+
+      if (!new Directory(currentConfig.directory).existsSync()) {
+        return new CliResult.fail(
+            'Must return a valid directory path, ${currentConfig.directory} is not a valid path');
+      }
+
+      if (currentConfig.react && currentConfig.env == Environment.vm) {
+        return new CliResult.fail('React tests can not be run on the vm');
```
evanweible-wf: This isn't necessarily true, is it? It is possible to run react on the server

lib/src/tasks/gen_test_runner/api.dart
```diff
@@ -0,0 +1,139 @@
...
+import 'package:dart_dev/src/tasks/task.dart';
+
+Future<GenTestRunnerTask> genTestRunner(
+    {String directory,
+    Environment environment,
+    String filename,
+    bool genHtml,
+    bool react,
+    List<String> scriptTags}) async {
+  var executable = 'gen-test-runner';
```
evanweible-wf: Where does this executable come from?

lib/src/tasks/gen_test_runner/api.dart
```diff
@@ -0,0 +1,139 @@
...
+  File generatedRunner = new File('$directory/$filename.dart');
+  IOSink writer = generatedRunner.openWrite(mode: FileMode.WRITE);
+
+  Directory testDirectory = new Directory(directory);
+  List<File> testFiles = [];
+  List<FileSystemEntity> allFiles =
+      testDirectory.listSync(recursive: true, followLinks: false);
+  allFiles.forEach((FileSystemEntity entity) {
+    if (entity is File) {
+      if (entity.path.contains(new RegExp(r'_test\.dart$'))) {
```
evanweible-wf:
```dart
if (entity.path.endsWith('_test.dart')) {
```

lib/src/tasks/gen_test_runner/api.dart
```diff
@@ -0,0 +1,139 @@
...
+        testFiles.add(entity);
+        task.testFiles.add(entity.path);
+      } else if (!entity.path.endsWith('$filename.dart') &&
+          entity.path.contains(new RegExp(r'\.dart$'))) {
```
evanweible-wf:
```
entity.path.endsWith('.dart')
```

lib/src/tasks/gen_test_runner/api.dart
```diff
@@ -0,0 +1,139 @@
...
+
+import 'package:dart_dev/src/tasks/gen_test_runner/config.dart';
+import 'package:dart_dev/src/tasks/task.dart';
+
+Future<GenTestRunnerTask> genTestRunner(
+    {String directory,
+    Environment environment,
+    String filename,
+    bool genHtml,
+    bool react,
```
evanweible-wf: I'm a little iffy about including this `react` flag in the API - it makes quite a few assumptions and is pretty specific to our uses. It assumes client-side react, but server-side react is valid. It's also tied to the `react_client.dart` import as well as the `setClientConfiguration()` api - if either of those were to change, it would put this task in the awkward position of having to support old and new.

I think we could facilitate the usage of react since it will be common for us without coupling it so closely to this API by providing a ReactClientHelper constant (that could be versioned to handle breaking changes) or something that would serve the purpose of injecting the react import and the configuration call.

test/integration/generated_runner.dart
evanweible-wf: okay, should clean up the remnants then

lib/src/tasks/gen_test_runner/api.dart
evanweible-wf: Sounds like this doesn't exist anymore and can be cleaned up

evanweible-wf:
@jayudey-wf first pass done - overall looks great! This will be really handy to have

lib/src/tasks/gen_test_runner/cli.dart
jayudey-wf: i don't know if that would work in this sense because i'm iterating over the results of the tasks that were generated from the configs rather then the configs themselvs

lib/src/tasks/gen_test_runner/cli.dart
jayudey-wf: this controls whether or not the reactClient will be set on the testRunner page https://github.com/Workiva/dart_dev/pull/124/files#diff-2c0d16a6e30daa966f4f73d0e7faf79bR87 so as currently implemented it could not be run on the vm

lib/src/tasks/gen_test_runner/api.dart
jayudey-wf: i kind of like keeping this in the task so it's more easily discerned which command generated which output when running the task

test/integration/gen_test_runner_test.dart
jayudey-wf: i updated the boolean to be a named parameter, i like keeping the method this way to reduce the number of method calls in the test bodies

lib/src/tasks/gen_test_runner/api.dart
evanweible-wf: I guess I'm confused then, didn't you say here that this is no longer a standalone executable: https://github.com/Workiva/dart_dev/pull/124/files#r51058202

The `GenTestRunnerTask` below doesn't do anything with the generated executable command - just stores it and outputs it later I assume. I'm fine with keeping the args so that they can be logged later, but building an executable command out of them that doesn't actually work is misleading

lib/src/tasks/gen_test_runner/cli.dart
evanweible-wf:
```dart
for (var i = 0; i < config.genTestRunner.configs.length; i++) {
  SingleRunnerConfig currentConfig = config.genTestRunner.configs[i];
```

```dart
for (var currentConfig in config.genTestRunner.configs) {
```

^ Those are equivalent

lib/src/tasks/gen_test_runner/cli.dart
evanweible-wf: Right, but react _could_ be run on the serve. This assumes that react is being run on the client. This goes back to comment from above about react (https://github.com/Workiva/dart_dev/pull/124/files#r51060811),

lib/src/tasks/gen_test_runner/cli.dart
jayudey-wf: sorry, i thought this was referencing something further down in the code, will update

lib/src/tasks/gen_test_runner/api.dart
jayudey-wf: ok, i'll just re-name it to be something like `taskArguments` instead?

evanweible-wf:
+1, and I can QA this when it gets to that point

README.md
```diff
@@ -409,6 +413,88 @@ object.
...
+        <tr>
+            <th>Name</th>
+            <th>Type</th>
+            <th>Default</th>
+            <th>Description</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td><code>additionalImports</code></td>
```
trentgrover-wf: seems odd to have additional paths / include files specified with different naming patterns (`additionalImports` + `scriptPaths`).  Should we pick 1 pattern or the other (`additionalImports + additionalScripts` or `importPaths + scriptPaths`)?

Personally, I'd be fine with `imports` and `scripts`.  Any other opinions?

README.md
```diff
@@ -409,6 +413,88 @@ object.
...
+            <td><code>List&lt;String&gt;</code></td>
+            <td><code>[]</code></td>
+            <td>Any other imports necessary to have in the generated runner file</td>
+        </tr>      
+        <tr>
+            <td><code>commandsPriorToTesting</code></td>
```
trentgrover-wf: not sure I like this naming.  how about `preTestCommands`?  any other suggestions?

lib/src/tasks/gen_test_runner/api.dart
```diff
@@ -0,0 +1,130 @@
+// Copyright 2015 Workiva Inc.
```
trentgrover-wf: we've hit 2016.  going to have to update all our copyrights :(

lib/src/tasks/gen_test_runner/api.dart
```diff
@@ -0,0 +1,130 @@
...
+Future testHtmlFileGenerator(
+    String directory, String filename, List<String> scriptPaths) async {
+  File generatedRunner = new File('$directory/$filename.html');
+  IOSink writer = generatedRunner.openWrite(mode: FileMode.WRITE);
+  writer.writeln('<!DOCTYPE html>');
+  writer.writeln('<html>');
+  writer.writeln('  <head>');
+  writer.writeln('    <title>$filename</title>');
+  scriptPaths.forEach((path) {
+    writer.writeln('    <script src="$path"></script>');
```
trentgrover-wf: there's a discrepancy in the usage between imports and scripts.  imports appear to expect that the user includes the full import line (`import 'package:fluri/fluri.dart';`), which seems reasonable since that will allow users to include things like `... as whatever;`, but the scripts expect only the script source path and infer the rest of the line.

what would you guys think about just using something like `dartHeaders` and `htmlHeaders` and expecting the full line in both places?  this would allow the dart headers to include things beyond imports (maybe annotation lines?) and the html headers to include things beyond scripts (maybe css?).  Can't think of a slew of use cases for this off the top of my head, but it's worth considering

trentgrover-wf:
code looks good @jayudey-wf , just want to get some other opinions on the api

README.md
evanweible-wf: yeah I'd be fine with `imports` and `scripts`

lib/src/tasks/gen_test_runner/api.dart
jayudey-wf: i can update that in a separate PR, that would touch a lot more files

lib/src/tasks/gen_test_runner/api.dart
evanweible-wf: Yeah this might make the most sense. I'd be fine with this

trentgrover-wf:
+1

jayudey-wf:
the one travis build is failing due to master not being merged in, failure is not related to this work https://travis-ci.org/Workiva/dart_dev/builds/106823597#L709

jayudey-wf:
@evanweible-wf this is ready for QA

evanweible-wf:
@jayudey-wf could you merge master in or rebase against latest master so that the push CI build doesn't fail?

lib/src/tasks/gen_test_runner/api.dart
```diff
@@ -0,0 +1,131 @@
...
+  IOSink writer = generatedRunner.openWrite(mode: FileMode.WRITE);
+
+  Directory testDirectory = new Directory(currentConfig.directory);
+  List<File> testFiles = [];
+  List<FileSystemEntity> allFiles =
+      testDirectory.listSync(recursive: true, followLinks: false);
+  allFiles.forEach((FileSystemEntity entity) {
+    if (entity is File) {
+      if (entity.path.endsWith('_test.dart') &&
+          !entity.path.endsWith('${currentConfig.filename}.dart')) {
```
trentgrover-wf: nit - since this condition is used twice, might as well do it once and hold the result in a bool for use in the 2 if statements

trentgrover-wf:
+1

johnbland-wf:
Just ran across this and thought I'd add one small suggestion not to hold up progress on this going through.

Naming the task `gen-test-runner` is very specific and limited. Calling it `generate` with an option of `test-runner` makes it available to other generators in the future.

trentgrover-wf:
good point @johnbland-wf , I like the consumption syntax of `ddev gen test-runner` + `ddev gen sauce-runner`, etc

I'll leave it up to @jayudey-wf whether this is a refactor worth doing as part of this PR or as a future followup

johnbland-wf:
Yup. Could even allow basic shells for tests, mixins, classes, etc.

evanweible-wf:
Sounds like https://github.com/Workiva/hoist

jayudey-wf:
I think I might defer that to a later change

johnbland-wf:
So many libraries, @evanweible-wf. -_-

jayudey-wf:
@dustinlessard-wf @evanweible-wf @maxwellpeterson-wf @trentgrover-wf the failing build issues have been resolved so I think this is ready for review again

dustinlessard-wf:
+1

maxwellpeterson-wf:
+1

evanweible-wf:
+1

jayudey-wf:
could someone +10 this?

evanweible-wf:
@jayudey-wf I will test this

evanweible-wf:
@jayudey-wf I didn't catch this in the code review - but this only supports creating a test file that must be run in a browser. This is inefficient for tests that _could_ be run in the VM, but more importantly it just doesn't work for tests that _must_ be run in the VM. It would be nice if the task could group the individual tests by their `@TestOn` annotations and generate a test runner for each group.

jayudey-wf:
that situation should work, you would just have to ensure that the config was set up correctly, https://github.com/Workiva/dart_dev/pull/124/files#diff-2c0d16a6e30daa966f4f73d0e7faf79bR51

evanweible-wf:
@jayudey-wf 

- If the `TestRunnerConfig.directory` ends in a slash, the pathing gets messed up:

    ```
    library test.unit..generated_runner;
    
    import '.browser_unit_test.dart' as browser_unit_test;
    import '.http/browser_utils_test.dart' as browser_utils_test;
    import '.http/client_test.dart' as client_test;
    ```

- If a test directory has duplicate file names across separate sub-directories, the import prefixing collides:

    ```
    library test.generated_runner;
    
    import './integration/http/client/browser_test.dart' as browser_test;
    import './integration/http/client/vm_test.dart' as vm_test;
    import './integration/http/common_request/browser_test.dart' as browser_test;
    import './integration/http/common_request/vm_test.dart' as vm_test;
    ```

    The prefixing may need to take into account the full path in order to avoid these collisions.

- The task detects the generated file as a non-test file and outputs a `[NOTICE]` about it. I'm assuming this is because the generated file is created before the test directory is scraped for test files. Not a huge deal, but does produce an unexpected notice.

    ```
    Generated using gen-test-runner -d test -e Environment.vm --genHtml
    [NOTICE] Found non-test dart file: test/generated_runner.dart
    ```

trentgrover-wf:
@evanweible-wf can you verify whether or not the last commit addresses all of your issues?

evanweible-wf:
## QA / +10

- [x] `TestRunnerConfig.directory` with a trailing slash no longer causes issues with the import pathing

- [x] Directory with duplicate file names across separate sub-directories no longer produces collisions

- [x] The generated test file is no longer being detected as a non-test file

- [x] Default configuration generates a test file with all tests imported for the browser environment

- [x] `TestRunnerConfig.dartHeaders` inserts lines after imports

- [x] `TestRunnerConfig.directory` defines the set of tests that are imported and run

- [x] `TestRunnerConfig.environment = Environment.vm` generates a file with `@TestOn('vm')`

- [x] `TestRunnerConfig.filename` changes the generated test file name

- [x] `TestRunnerConfig.genHtml` dictates whether or not the companion HTML file is generated

- [x] `TestRunnerConfig.htmlHeaders` inserts lines in the `<head>`

- [x] `TestRunnerConfig.preTestCommands` inserts commands at the beginning of the `main()` block

evanweible-wf:
@jayudey-wf ready for merge

maxwellpeterson-wf:
QA +1

Merging into master.

maxwellpeterson-wf:
+1

evanweible-wf:
Ran through the manual CI steps with the last two commits included:

#### Manual CI

- [x] `ddev format` passes locally

- [x] `ddev analyze` passes locally

- [x] `ddev test --integration` all 45 tests pass locally

trentgrover-wf:
+1

maxwellpeterson-wf:
QA +1

Merging into master.

Issue #55, opened 2015-08-27
CP-2142 Test task: ability to run a subset of the tests


maxwellpeterson-wf:
I want to be able to run a subset of the tests:

``` bash
$ pub run dart_dev test test/my_module/

$ pub run dart_dev test test/my_module/test_module.dart
```

evanweible-wf:
+1, this can be supported. Also, I want to add support for the test packages `-n` flag which allows you to pass in a string that gets matched against the test names and only matching tests run

trentgrover-wf:
+1 from Tim McCall

markerickson-wf:
+1

maxwellpeterson-wf:
Addressed.

PR #126, opened 2016-01-29
CP-1318 Add ability to pass in test name as a parameter to test task

jayudey-wf:
## Issue
- https://github.com/Workiva/dart_dev/issues/55 (master currently address part of this by a user being able to add in individual test files)

## Changes
**Source:**
- added the ability to pass in a test name through the CLI
- updated analyze test due to changes in dart 1.14.0

**Tests:**
- added

## Areas of Regression
- test execution

## Testing
- `ddev test` works as expected with and without a named test

## Code Review
@trentgrover-wf
@maxwellpeterson-wf
@evanweible-wf
@dustinlessard-wf

### DIFF
--- a/lib/src/tasks/test/cli.dart
+++ b/lib/src/tasks/test/cli.dart
@@ -89,8 +89,7 @@ class TestCli extends TaskCli {
 
     bool unit = parsedArgs['unit'];
     bool integration = parsedArgs['integration'];
-    bool testNamed;
-    parsedArgs['name'] != null ? testNamed = true : '';
+    bool testNamed = parsedArgs['name'] != null;
     List<String> tests = [];
     int individualTests = 0;
 
--- a/lib/src/tasks/test/cli.dart
+++ b/lib/src/tasks/test/cli.dart
@@ -47,7 +47,7 @@ class TestCli extends TaskCli {
         allowMultiple: true,
         help:
             'The platform(s) on which to run the tests.\n[vm (default), dartium, content-shell, chrome, phantomjs, firefox, safari]')
-    ..addOption('test-name',
+    ..addOption('name',
         abbr: 'n',
         help:
             'A substring of the name of the test to run.\nRegular expression syntax is supported.');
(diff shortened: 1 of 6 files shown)


lib/src/tasks/test/cli.dart
```diff
@@ -46,7 +46,11 @@ class TestCli extends TaskCli {
         abbr: 'p',
         allowMultiple: true,
         help:
-            'The platform(s) on which to run the tests.\n[vm (default), dartium, content-shell, chrome, phantomjs, firefox, safari]');
+            'The platform(s) on which to run the tests.\n[vm (default), dartium, content-shell, chrome, phantomjs, firefox, safari]')
+    ..addOption('test-name',
+        abbr: 'n',
```
evanweible-wf: We should be consistent with the `test` package, and they name the option `name`:

```
-n, --name                     A substring of the name of the test to run.
                               Regular expression syntax is supported.
```

evanweible-wf:
Just the one note, then this looks good!

dustinlessard-wf:
Awesome!

jayudey-wf:
+10

- intalled version into web_skin_dart, verified that `test` task worked as expected without a named parameter and that tests were able to run individually and as a group with the named parameter option

lib/src/tasks/test/cli.dart
```diff
@@ -85,6 +89,8 @@ class TestCli extends TaskCli {
 
     bool unit = parsedArgs['unit'];
     bool integration = parsedArgs['integration'];
+    bool testNamed;
+    parsedArgs['name'] != null ? testNamed = true : '';
```
evanweible-wf: This syntax is a bit weird. Ternary is usually used like so:
```dart
bool testNamed = parsedArgs['name'] != null ? [value if true] : [value if false]
```
In this case, it's just a bool, so it could be:
```dart
bool testNamed = parsedArgs['name'] != null;
```
Another thing to note - as it is currently written, `testNamed` will either be `true` or `null`, which makes me think the if statement below on line 167 could throw (null boolean).

evanweible-wf:
+1

trentgrover-wf:
+1
@evanweible-wf how much work will it be to incorporate your lazy args spike?  that would eliminate the need for this sort of specificity, correct?

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - passing CI, installed branch into web_skin_dart and verified that tests were still running as expected with or without named tests
- [ ] All unit tests pass 

Merging into master.

charliekump-wf:
+10 for posterity. CI build successfully running dart_dev test.

PR #141, opened 2016-03-22
CP-1480 Test Task: correctly build the set of tests to run and respect the --no-color flag

### LINKED ISSUE #137: Running test with "--no-color" doesn't remove all coloring from output
When running tests with "pub run dart_dev test --no-color" with tests that use pub serve, the color is not removed from the output of the subprocess call "pub run test ...". It is removed from the output of the "pub serve..." call.

image

### LINKED ISSUE #138: Error thrown when attempting to run Dart_Dev Test with specific test file path included
The readme states "Individual test files can be executed by appending their path to the end of the command." but it appears this may not be working anymore as I get the following error when I attempt to do so.  The test passes if run by name or as part of the whole package, so it isn't an issue with the test itself.

Image of error
http://screencast.com/t/WiKRrhC0c

evanweible-wf:
Fix #137.
Fix #138.

## Issues
Originally discovered in #138. The test is incorrectly including the default unit test suite even when individual tests are given as args and even with `--no-unit`.

Additionally, the test task does not respect the `--no-color` flag (#137).

## Changes
- Update the `TaskCli.run` abstract method to take `color` as an optional param. This allows the top-level CLI to pass the value of the `color` flag into the tasks.

- Forward the `--no-color` flag onto the the `pub run test` command.

- Update the logic used to build the list of tests to run as follows:
  - If individual test files/directories are passed in, only those are run.
  - Unless `--no-unit` is passed in, run the unit test suite.
  - If `--integration` is passed in, run the integration suite.

- The tests for the test task have been updated such that tests that expect the test task to succeed now also verify the number of tests that were run.

## Testing
- CI passes
- Checkout 0dc911c and verify that the newly updated tests that expect only a certain number of tests to run fail because of the issue from #138 (`ddev test --integration -n "Test Task"`)
- Checkout 20c8037 and verify that the test task tests pass (`ddev test --integration -n "Test Task"`)

## Code Review
@trentgrover-wf 
@maxwellpeterson-wf 
@dustinlessard-wf 
@jayudey-wf 

fyi: @trentonsmith-wf @philipbalvanz-wf @johnbland-wf @mattsanders-wf

### DIFF
--- a/lib/src/tasks/test/cli.dart
+++ b/lib/src/tasks/test/cli.dart
@@ -15,7 +15,6 @@
 library dart_dev.src.tasks.test.cli;
 
 import 'dart:async';
-import 'dart:io';
 
 import 'package:args/args.dart';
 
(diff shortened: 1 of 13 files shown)


johnbland-wf:
Looks good. Thx for fast tracking this.

maxwellpeterson-wf:
+1

trentgrover-wf:
+1
@jayudey-wf ready for merge

lib/src/tasks/test/cli.dart
```diff
@@ -104,20 +80,37 @@ class TestCli extends TaskCli {
...
+      // were not specified.
+      if (unit) {
+        tests.addAll(config.test.unitTests);
+      }
+      if (integration) {
+        tests.addAll(config.test.integrationTests);
+      }
     }
 
     if (tests.isEmpty) {
```
jayudey-wf: do we want to move this conditional up into the block above because currently you can see a failure if you try to run a named test but don't have unit tests specified in your dev.dart file

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
- [ ] Dev/QA +10 with detail of what was tested'
 - verified that when passing in a test file name that only the tests from within that test file are run
 - verified that the `--no-color` flag was being appropriately piped to test tasks using pub-serve
- [ ] All unit tests pass 

Merging into master.

PR #151, opened 2016-05-03
CP-1794 Add completions.dart support

georgelesica-wf:
This addresses #103. I've added basic support for generating the completion file using completions.dart.

See the README for information on how to consume the new completions.

The script even works with custom project-specific tasks because it operates by adding a pseudo command to ddev that effectively introspects its current configuration. This means the script won't actually need updating all that often. In the future we may not even need to store the script, we may just generate it on the fly from completions.dart.

## Code Review / +10

Follow the instructions in the README to get it working. Then make sure it works.

Install globally using `pub global activate -s /path/to/clone` instead of the command in the README since that would just install the release version. Of particular use would be testing in zsh since I don't use it.

The completions only work within a project, see the readme for how to set up the `ddev` alias, which is required. Link this branch into a random Dart project, pub upgrade, and then you should have completions when you do `ddev <tab> <tab> <tab>...`.

## Review:

@evanweible-wf 
@maxwellpeterson-wf 
@jayudey-wf

### DIFF
--- a/lib/src/dart_dev_cli.dart
+++ b/lib/src/dart_dev_cli.dart
@@ -18,7 +18,7 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:args/args.dart';
-
+import 'package:completion/completion.dart';
 import 'package:dart_dev/util.dart'
     show
         TaskProcess,
@@ -101,7 +101,7 @@ Future _run(List<String> args) async {
 
   ArgResults env;
   try {
-    env = _parser.parse(args);
+    env = tryArgsCompletion(args, _parser);
   } on FormatException catch (e) {
     reporter.error('${e.message}\n', shout: true);
     reporter.log(_generateUsage(), shout: true);
--- a/lib/src/dart_dev_cli.dart
+++ b/lib/src/dart_dev_cli.dart
@@ -30,6 +30,7 @@ import 'package:dart_dev/src/tasks/cli.dart';
 import 'package:dart_dev/src/tasks/config.dart';
 
 import 'package:dart_dev/src/tasks/analyze/cli.dart';
+import 'package:dart_dev/src/tasks/bash_completion/cli.dart';
 import 'package:dart_dev/src/tasks/copy_license/cli.dart';
 import 'package:dart_dev/src/tasks/coverage/cli.dart';
 import 'package:dart_dev/src/tasks/docs/cli.dart';
@@ -55,6 +56,7 @@ String _topLevelUsage = _parser.usage;
 
 dev(List<String> args) async {
   registerTask(new AnalyzeCli(), config.analyze);
+  registerTask(new BashCompletionCli(), config.bashCompletion);
   registerTask(new CopyLicenseCli(), config.copyLicense);
   registerTask(new CoverageCli(), config.coverage);
   registerTask(new DocsCli(), config.docs);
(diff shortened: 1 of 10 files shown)


lib/src/dart_dev_cli.dart
```diff
@@ -101,7 +101,7 @@ Future _run(List<String> args) async {
 
   ArgResults env;
   try {
-    env = _parser.parse(args);
+    env = tryArgsCompletion(args, _parser);
```
evanweible-wf: So does this just try add support for parsing incomplete args? Like if someone were to accidentally run `ddev form` - it would run `ddev format`?

georgelesica-wf: I didn't read through the full source, but it seems like it basically just adds a new command that provides completions. The completion script is then almost entirely generic in that it just runs a special version of ddev itself to get the completions.

evanweible-wf: Ah ok I see 👍

georgelesica-wf:
The one drawback here is that the completions are pretty slow because dart_dev essentially has to run every time it needs to do a completion. Which isn't great since there is quite a bit of "stuff" done at startup.

evanweible-wf:
@georgelesica-wf just pulled it down to test it out, it is pretty slow.. is there anyway we could provide an alternate code path for completion to speed it up?

georgelesica-wf:
@evanweible-wf I'll look into it, I agree the (lack of) speed is annoying.

georgelesica-wf:
@evanweible-wf upon investigation, it is already short-circuiting for us (https://github.com/kevmoo/completion.dart/blob/master/lib/src/try_completion.dart#L69) so I was wrong about why it is slow. The only setup we're doing before it takes control is registering all our commands with the arg parser, which is necessary. I'm not entirely sure, now, why completion is so slow.

evanweible-wf:
@georgelesica-wf we can revisit performance at a later time - this is definitely the more maintainable approach so I'm good with it.

+1

trentgrover-wf:
+1

maxwellpeterson-wf:
+1

evanweible-wf:
@jayudey-wf ready for merge

georgelesica-wf:
@jayudey-wf I updated the PR and rebuilt the completion script so it should be unambiguous now as to when the completions will work. Spoiler: only in a project, not globally.

jayudey-wf:
+5 i've verified that it's working as expected with bash but I am currently un-able to verify it's working as expected within zsh

evanweible-wf:
@jayudey-wf I'll test with zsh

evanweible-wf:
+5 
- [x] completions work with zsh when setup according to the README instructions

@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - performed and documented by Evan and Jay
- [ ] All unit tests pass

Merging into master.

evanweible-wf:
+1

jayudey-wf:
Manual CI on dart 1.15.0

- [x] `pub run dart_dev format`
- [x] `pub run dart_dev analyze`
- [x] `pub run dart_dev test --integration`
- [x] `pub run dart_dev coverage --integration`

PR #152, opened 2016-05-06
Fix Firefox test failures in Sauce Labs due to window not being active

greglittlefield-wf:
## Ultimate Problem
For some reason, Sauce Labs no longer activates the Firefox window when running tests. In Firefox, if the window isn't active, focus events won't be fired (see <https://bugzilla.mozilla.org/show_bug.cgi?id=566671>).

This causes focus-related unit tests to fail in Firefox for Sauce Labs builds.

## Solution
_Focusing the iframe, its window and then its body via JS doesn't work._

_Focusing/blurring all kinds of other elements via JS doesn't work._

_Setting the iframe as the active frame in the WebDriver doesn't work._

What does work? Having the WebDriver click the page before it loads our tests, which has the side effect of activating the browser window.

## Testing
Verify that focus-related tests pass. (see https://github.com/Workiva/web_skin_dart/pull/391)

## Areas of Regression
Sauce Labs builds.

---

FYA: @jayudey-wf 
FYI: @joelleibow-wf @jacehensley-wf @clairesarsam-wf @aaronlademann-wf

### DIFF
--- a/lib/src/tasks/saucelabs_tests/sauce_unit_test_driver.dart
+++ b/lib/src/tasks/saucelabs_tests/sauce_unit_test_driver.dart
@@ -6,7 +6,7 @@ import 'dart:io';
 
 import 'package:dart_dev/util.dart';
 import 'package:webdriver/io.dart'
-    show BrowserPlatform, Capabilities, WebDriver, createDriver;
+    show By, Capabilities, WebDriver, createDriver;
 
 import './platforms.dart';
 import 'sauce_runner.dart';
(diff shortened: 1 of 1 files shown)


lib/src/tasks/saucelabs_tests/sauce_unit_test_driver.dart
```diff
@@ -157,6 +157,18 @@ class SauceUnitTestDriver {
...
+      // If the browser window is not active, focus events won't be dispatched in Firefox as expected.
+      //
+      // So, before running any tests, we'll ensure the browser window is active by clicking on the page with WebDriver.
+      //
+      // 1. Load `about:blank` to ensure that clicking on the page does nothing.
+      // 2. Move the cursor over a valid element so that the WebDriver doesn't break itself trying to click on nothing.
+      // 3. Click.
+      await _driver.get('about:blank'); // [1]
+      await _driver.mouse.moveTo(
+          element: await _driver.findElement(const By.tagName('body'))); // [2]
```
jayudey-wf: do you also need to add `By` to the show clause coming from `'package:webdriver/io.dart'`

greglittlefield-wf: Ah, dang; yes. Sorry, copy/pasted without seeing whether it analyzed properly. Will fix

greglittlefield-wf: Fixed.

jayudey-wf:
+1, merging into branch

PR #195, opened 2016-12-06
CP-2124: Add Debug Flag to Test Task

### LINKED ISSUE #120: CP-2124 Better test debugging: support `--pause-after-load`
The `test` package provides a `--pause-after-load` flag (see https://github.com/dart-lang/test#debugging) that pauses the tests before they run, providing an opportunity to open the debugger.

Consumers of `dart_dev` should be able to take advantage of this option somehow.

Perhaps the `dart_dev test` command's CLI could be refactored to forward any extra arguments to the `test` package's CLI?

timmccall-wf:
Fixes #120 

Looking for some initial feedback regarding approach, etc.

@evanweible-wf 

## Testing

- [ ] CI passes

FYI @greglittlefield-wf @patkujawa-wf

### DIFF
--- a/lib/src/tasks/test/cli.dart
+++ b/lib/src/tasks/test/cli.dart
@@ -44,10 +44,10 @@ class TestCli extends TaskCli {
         negatable: true,
         defaultsTo: defaultPubServe,
         help: 'Serves browser tests using a Pub server.')
-    ..addFlag('debug',
-        abbr: 'd',
+    ..addFlag('pause-after-load',
+        abbr: 'w',
         negatable: true,
-        defaultsTo: defaultDebug,
+        defaultsTo: defaultPauseAfterLoad,
         help: 'Debug browser tests using the "--pause-after-load" flag.')
     ..addOption('pub-serve-port',
         help:
--- a/lib/src/tasks/test/cli.dart
+++ b/lib/src/tasks/test/cli.dart
@@ -45,10 +45,10 @@ class TestCli extends TaskCli {
         defaultsTo: defaultPubServe,
         help: 'Serves browser tests using a Pub server.')
     ..addFlag('pause-after-load',
-        abbr: 'w',
-        negatable: true,
-        defaultsTo: defaultPauseAfterLoad,
-        help: 'Debug browser tests using the "--pause-after-load" flag.')
+        help: 'Pauses for debugging before any tests execute.\n'
+            'Implies --concurrency=1 and --timeout=none.\n'
+            'Currently only supported for browser tests.',
+        negatable: false)
     ..addOption('pub-serve-port',
         help:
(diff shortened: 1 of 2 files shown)


lib/src/tasks/test/cli.dart
```diff
@@ -44,6 +44,11 @@ class TestCli extends TaskCli {
         negatable: true,
         defaultsTo: defaultPubServe,
         help: 'Serves browser tests using a Pub server.')
+    ..addFlag('debug',
+        abbr: 'd',
+        negatable: true,
+        defaultsTo: defaultDebug,
+        help: 'Debug browser tests using the "--pause-after-load" flag.')
```
evanweible-wf: If this set more than one flag, I'd be fine with naming it `debug`, but since it's a 1:1 mapping I'd rather keep it consistent. Also, with v2, all unused flags/options will be passed through to the underlying executable, so it would need to be `--pause-after-load` at that point anyway.

evanweible-wf:
@timmccall-wf just the one comment

lib/src/tasks/test/cli.dart
timmccall-wf: Sounds good. Will amend.

maxwellpeterson-wf:
+1

lib/src/tasks/test/cli.dart
```diff
@@ -44,6 +44,11 @@ class TestCli extends TaskCli {
         negatable: true,
         defaultsTo: defaultPubServe,
         help: 'Serves browser tests using a Pub server.')
+    ..addFlag('pause-after-load',
+        abbr: 'w',
+        negatable: true,
```
evanweible-wf: The `--pause-after-load` flag from the `test` package has no abbreviation and is not negatable, mind if we stay consistent with that? (https://github.com/dart-lang/test/blob/17cd53ebf069b27e25d609aa431f6abe6bf0f68f/lib/src/runner/configuration/args.dart#L83-L87)

timmccall-wf: Not at all. Done.

evanweible-wf:
+1

evanweible-wf:
Thanks for the contribution @timmccall-wf!

patkujawa-wf:
CI failure seems intermittent, especially since the tests passed when run with coverage.

evanweible-wf:
@patkujawa-wf CI passed on a rebuild, we'll get this merged

maxwellpeterson-wf:
+1

evanweible-wf:
+10 
- [x] CI passes

evanweible-wf:
QA +1
- [x] Testing instruction
- [x] Dev +1's
- [x] Dev/QA +10 with detail of what was tested
  - n/a - simple change, no test needed
- [x] All unit tests pass

Merging into master.

Issue #51, opened 2015-08-26
CP-2144 Format Task: specify list of files to format/check


evanweible-wf:
By default, the format task should use the project's configuration for files to format. But, we should also allow running the formatter on a specific file or set of files. This can be done by allowing additional args to the format CLI that override the default configuration.

trentgrover-wf:
+1 from George Lessica

PR #222, opened 2017-05-25
UIP-2338 / CP-2144: Support custom local format task + specifying files via CLI

greglittlefield-wf:
## Description
- There were several tweaks that needed to be made to support a custom local formatter task in dart_dev.

- https://github.com/Workiva/dart_dev/issues/51:
    > By default, the format task should use the project's configuration for files to format. But, we should also allow running the formatter on a specific file or set of files. This can be done by allowing additional args to the format CLI that override the default configuration.

    I also wanted this behavior for my custom formatter task and needed to make some changes in dart_dev to support it, so I thought it'd only be fair to address that issue while I was at it 😃.

## Solution
- Format task:
    - Split out and export logic for expanding lists of formatted files into `getFilesToFormat`, for use in custom format task
    - Speed up run time by:
    image
        - Not resolving absolute paths unnecessarily
        - Not following symlinks
            - __Potential breakage:__ this is a behavioral change, although in master, symlinks are only followed when there are excludes present.
     - Allow specifying specific directories/files via CLI rest args
    - Update formatter API to be able to take in specific files directly
    - Deprecate `FormatConfig.directories`, replacing it with and proxying it to `FormatConfig.paths` for clearer naming (since files can now be specified directly), and deprecate/proxy other `directories` named arguments accordingly
    - Update functional tests.
- Add new `TaskCLI.usage` getter that allows for custom usage strings, to allow for indicating that `format`/`test` package accept rest args
    ```diff
     $ ddev format --help
     Standardized tooling for Dart projects.

    -Usage: pub run dart_dev format [options]
    +Usage: pub run dart_dev format [options] [files or directories...]
    ```

- Add `ExportConfig` task (`ddev export-config`) to allow local tasks to read configuration (necessary since they're launched in separate processes)
    - This is a stopgap solution to support the new custom format task until dart_dev 2.0, which will hopefully have better configuration sharing 
    - `Reporer`: don't colorize empty strings (so that `[38;5;2m[0m` isn't appended to JSON output due to printing empty success result string)
- Allow overriding of built-in tasks (e.g., `format`)
- Update Regex to fix AnalyzerTask functional tests that were failing with the latest analyzer, due to CLI output changes

## Testing:
- Verify that functional tests pass
- Verify `ddev format --help` looks good
- Verify `ddev test --help` looks good

## Areas of Regression
- Format task
- Colorized output
- `--help` flag


lib/src/tasks/export_config/cli.dart
```diff
@@ -0,0 +1,42 @@
...
+  Future<CliResult> run(ArgResults parsedArgs, {bool color: true}) async {
+    reporter.log(new JsonEncoder.withIndent('  ', (obj) {
+      // Attempt to convert the config into JSON.
+      try {
+        return obj.toJson();
+      } catch (_) {}
+
+      // If the given object doesn't support JSON serialization, skip it.
+      // This allows us to not implement serialization in all configs up-front,
+      // as well as not break for parts of the config that can't be serialized.
```
greglittlefield-wf: `'format'` is the only one implemented; all others are `null`. Should I fill these in for consistency/completeness, or are they fine staying "unimplemented" until needed?

greglittlefield-wf:
Please review: @Workiva/web-platform-pp @Workiva/ui-platform-pp

lib/src/dart_dev_cli.dart
```diff
@@ -35,6 +35,7 @@ import 'package:dart_dev/src/tasks/copy_license/cli.dart';
 import 'package:dart_dev/src/tasks/coverage/cli.dart';
 import 'package:dart_dev/src/tasks/docs/cli.dart';
 import 'package:dart_dev/src/tasks/examples/cli.dart';
+import 'package:dart_dev/src/tasks/export_config//cli.dart';
```
evanweible-wf: double slash

lib/src/dart_dev_cli.dart
```diff
@@ -89,12 +91,6 @@ dev(List<String> args) async {
 }
 
 void registerTask(TaskCli cli, TaskConfig config) {
-  if (_cliTasks[cli.command] != null) {
-    throw new ArgumentError(
```
evanweible-wf: Should we print a warning about the fact that a task is being overridden?

lib/src/tasks/export_config/cli.dart
evanweible-wf: Sorry @greglittlefield-wf I accidentally deleted your comment instead of my reply..

> 'format' is the only one implemented; all others are null. Should I fill these in for consistency/completeness, or are they fine staying "unimplemented" until needed?

I'd say that's fine, especially since this is going to be an undocumented feature

lib/src/dart_dev_cli.dart
greglittlefield-wf: We could, but do we want that warning showing up every time dart_dev is run?

greglittlefield-wf: Mm, or perhaps we should not throw when a local task overrides a built-in one, and warn in all other cases?

greglittlefield-wf: Actually, due to the way that `LocalCli.discover` works, it's not possible to detect local tasks overriding other local tasks without overhauling how that works... any local tasks with the same name trump other ones.

What about something like this? https://github.com/Workiva/dart_dev/commit/6c029f3aaf6bb0f2c3ad36c279757e883b3979b0

greglittlefield-wf:
Updated description and testing instructions to include a solution for https://github.com/Workiva/dart_dev/issues/51

aaronlademann-wf:
+10

aaronlademann-wf:
Could we get a +1 from @Workiva/web-platform-pp folk(s)?

jayudey-wf:
will this end up being a breaking change since `directories` -> `paths`

greglittlefield-wf:
@jayudey-wf It should not; `directories` is deprecated but should still behave as expected, proxying `paths`. Tests around this can be found here: https://github.com/Workiva/dart_dev/pull/222/files#diff-507150868c16f0a55a3305b8d7748a43R250

trentgrover-wf:
@jayudey-wf ready for merge and release - unless @evanweible-wf wants a final look

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - performed and documented by Aaron
- [ ] Security review (if required)
- [ ] All unit tests pass
- [ ] Rosie has run and reports properly the release the ticket will be included in
- [ ] **Dependency Scan Clean**

Merging into master.

PR #231, opened 2017-07-18
WP-4746 Ensure task runner waits for stdout/stderr before considering task "done"

evanweible-wf:
## Issue

We are seeing the task-runner tests fail intermittently because the test task output is incomplete (it does not include the "All tests passed." or "Some tests failed." line that the test is looking for). I haven't been able to reproduce this locally.

## Solution

My best guess is that the task runner is considering the tasks done before the stdout/stderr streams have been completely processed due to it listening to the `exitCode` future instead of `done`.

Additionally, there were two dependencies that are being used but aren't explicitly depended on, so I've added those.

## Testing

- [ ] CI passes (maybe a couple of times?)
- [ ] `pub publish --dry-run` passes with no warnings

## Code Review

@Workiva/web-platform-pp

### DIFF
--- a/test_fixtures/task_runner/failing_tasks/test/simple_test.dart
+++ b/test_fixtures/task_runner/failing_tasks/test/simple_test.dart
@@ -3,8 +3,8 @@ import 'dart:async';
 import 'package:test/test.dart';
 
 void main() {
-  test('passes', () async {
+  test('fails', () async {
     await new Future.delayed(new Duration(seconds:5));
-    expect(true, isTrue);
+    expect(true, isFalse);
   });
 }
--- a/test_fixtures/task_runner/passing_tasks/test/simple_test.dart
+++ b/test_fixtures/task_runner/passing_tasks/test/simple_test.dart
@@ -1,7 +1,11 @@
+import 'dart:async';
+
 import 'package:test/test.dart';
 
 void main() {
-  test('passes', () {
+  test('passes', () async {
+    // Wait so that the test task doesn't complete too quickly.
+    await new Future.delayed(const Duration(seconds: 2));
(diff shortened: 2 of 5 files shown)


aviary2-wf:
## Raven
<details>
<summary>Number of Findings: 0</summary>

</details>

test_fixtures/task_runner/failing_tasks/test/simple_test.dart
```diff
@@ -5,6 +5,6 @@ import 'package:test/test.dart';
 void main() {
   test('passes', () async {
```
jayudey-wf: can you update the title of the test here?

test_fixtures/task_runner/passing_tasks/test/simple_test.dart
```diff
@@ -1,7 +1,10 @@
+import 'dart:async';
+
 import 'package:test/test.dart';
 
 void main() {
-  test('passes', () {
+  test('passes', () async {
+    await new Future.delayed(const Duration(seconds: 2));
```
jayudey-wf: could you add a comment as to why this is here? I am assuming it's so that the task runner kills the test task since it will complete after the format task fails

evanweible-wf:
@jayudey-wf ready for you

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - passing CI (multiple runs)
  - verified that `pub publish --dry-run` produces no warnings
- [ ] Security review (if required)
- [ ] All unit tests pass
- [ ] Rosie has run and reports properly the release the ticket will be included in

Merging into master.

PR #249, opened 2018-01-11
Generated dart runner file does not comply with directives_ordering lint

danaoredson-wf:
The dart file generated by `pub run dart_dev format` does not comply with the directives_ordering lint.

This is easy to fix, by simply importing 'package:test/test.dart' before the relative imports.


danaoredson-wf:
@jacehensley-wf @evanweible-wf @maxwellpeterson-wf @brianneal-wf can you review the last couple of commits?

PR #271, opened 2018-08-01
Update webdriver dep for Dart 2 compatibility

maxwellpeterson-wf:
We will need to be able to fetch `webdriver 2.x.x` when on Dart 2.

Expand acceptable dependency range on `webdriver` to include 2.0


sebastianmalysa-wf:
QA +1
- [x] Dev +1's
- [x] CI passes

PR #287, opened 2019-01-11
AF-3801 fix reporter runtime error in Dart 2.1.0

robbecker-wf:
# Overview
Under Dart 2.1.0, when running tests that use `reporter.log`. The `quiet` variable would be null somehow and cause a runtime error in the conditional check in the `_log` method. 
I imagine this is caused by something like checked mode when running tests. I could not reproduce this runtime error when running a similar thing on the command line (prod mode).

An example failing build due to this error: https://ci.webfilings.com/build/1585120

# Changes
- Don't allow these to variables to be null

# Testing
- CI passes
- See passing build using this branch: https://github.com/Workiva/over_react_format/pull/25

### DIFF
--- a/lib/src/reporter.dart
+++ b/lib/src/reporter.dart
@@ -93,7 +93,10 @@ class Reporter {
       color && message.isNotEmpty ? pen(message) : message;
 
   void _log(IOSink sink, String message, {bool shout: false}) {
-    if (quiet && !shout) return;
+    // This may look like a strange conditional, but please leave it.
+    // Somehow when using the tearoff `reporter.log` the `quiet` variable
+    // is null, which causes a runtime error when running tests in Dart 2.1.0
+    if (quiet == true && shout != true) return;
     sink.writeln(message);
   }
 }
--- a/lib/src/reporter.dart
+++ b/lib/src/reporter.dart
@@ -30,7 +30,11 @@ class Reporter {
   bool color = true;
   bool quiet = false;
 
-  Reporter({bool this.color, bool this.quiet});
+  Reporter({this.color: true, this.quiet: false}) {
+    // Do an additional check here to make sure nulls were not passed
+    color ??= true;
+    quiet ??= false;
+  }
 
   String colorBlue(String message) => _color(_blue, message);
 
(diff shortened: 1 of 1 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/reporter.dart
```diff
@@ -93,7 +93,10 @@ class Reporter {
       color && message.isNotEmpty ? pen(message) : message;
 
   void _log(IOSink sink, String message, {bool shout: false}) {
-    if (quiet && !shout) return;
+    // This may look like a strange conditional, but please leave it.
+    // Somehow when using the tearoff `reporter.log` the `quiet` variable
+    // is null, which causes a runtime error when running tests in Dart 2.1.0
+    if (quiet == true && shout != true) return;
```
greglittlefield-wf: Could it be that `Reporter` isn't getting passed both arguments during intiialization?
https://github.com/Workiva/dart_dev/blob/7eee6da5affe23b2f9cd1b8241a18431aafaff7c/lib/src/reporter.dart#L22

Based on the existing code, 
https://github.com/Workiva/dart_dev/blob/a3b36d0b7be50338751d421cf3c3813f4475f1c5/lib/src/reporter.dart#L30-L35
, I would expect `new Reporter()` to have both fields be null.

I feel like it should just be:
```dart
bool color;
bool quiet;

Reporter({bool this.color = true, bool this.quiet = false}); 
```

Could it be that it's only showing up in testing because this is being run in checked mode?

lib/src/reporter.dart
```diff
@@ -30,7 +30,11 @@ class Reporter {
   bool color = true;
   bool quiet = false;
 
-  Reporter({bool this.color, bool this.quiet});
+  Reporter({this.color: true, this.quiet: false}) {
```
greglittlefield-wf: #nit these defaults aren't necessary if we have null-awares below

lib/src/reporter.dart
```diff
@@ -30,7 +30,11 @@ class Reporter {
   bool color = true;
   bool quiet = false;
```
greglittlefield-wf: #nit these defaults aren't necessary

lib/src/reporter.dart
```diff
@@ -93,7 +97,11 @@ class Reporter {
       color && message.isNotEmpty ? pen(message) : message;
 
   void _log(IOSink sink, String message, {bool shout: false}) {
-    if (quiet && !shout) return;
+    // Ensure that even if quiet or shout are null, the conditional
+    // will result in a boolean
+    if (quiet == true && shout != true) {
```
greglittlefield-wf: I don't think we should have to worry about this case; these should never be null unless someone explicitly sets them to null, in which case the program should probably fail.

lib/src/reporter.dart
robbecker-wf: I left them in so that when calling the constructor, the IDE would tell you what the defaults are.

lib/src/reporter.dart
robbecker-wf: Totally. I'll remove them.

evanweible-wf:
QA +1
- [x] CI passes

@Workiva/release-management-p

PR #291, opened 2019-03-08
CPLAT-4822: Fail test task if `build_runner build` fails

evanweible-wf:
# [CPLAT-4822](https://jira.atl.workiva.net/browse/CPLAT-4822)
Issue Status

## Problem

On Dart 2, the `test` task currently runs tests via `build_runner test`, but unfortunately this command does not exit with a non-zero exit code if the build fails, which means that the dart_dev `test` task can appear as though it passed when in fact no tests were even run.

## Solution

When necessary, run `build runner build` first and fail early if it exits with a non-zero exit code. If it succeeds, then we can proceed to run the tests via `build_runner test`, knowing that they will actually run.

## Testing

- [ ] CI passes
- [ ] Verify this behavior against the latest react_tracing dart2 branch, which is where we discovered this issue

## Code Review
@maxwellpeterson-wf @smaifullerton-wk @seanburke-wf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Changelog
 
+## [2.0.4](https://github.com/Workiva/dart_dev/compare/2.0.3...2.0.4)
+
+- **Bug Fix:** When on Dart 2 and the package has a dependency on `build_test`,
+  the `test` task will now properly exit with a non-zero exit code if the build
+  fails.
+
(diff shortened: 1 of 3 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

maxwellpeterson-wf:
QA +1
- [x] CI passes
- [x] Verified in react_tracing that this branch properly fails early when running `test` task and the build fails
- [x] Verified in a project with a working build that it correctly runs tests after the build
- [x] Verified on Dart 1 that no build is run at all

PR #297, opened 2019-05-11
TaskProcess: kill when signal is received, killing fixes/improvements

greglittlefield-wf:
### Motivation
Sometimes, custom tasks would remain running in the background when dart_dev was sent a signal (e.g., `^C`).

This happened especially often with our `ddev serve`.

### Changes
- Automatically kill child processes and all their descendants when SIGINT or SIGTERM is received
- Fix bug in `killAllChildren` that, if a single child didn't exit, prevented subsequent children from exiting
    - The offending code: ` killed = killed && Process.killPid(cpids[i], signal);`
- Deprecate `killAllChildren` and replace it with a synchronous implementation with a more accurate name, `killAllDescendants`
    - I switched this to a synchronous implementation to reduce the risk of race conditions, and allow it to be called reliably from synchronous functions
- Add `killProcessAndAllDescendants` convenience method

### Testing
TODO


aviary3-wk:
## Security Insights

## Action Items
---

evanweible-wf:
@greglittlefield-wf In the v3 branch I've been using a util that waits for and ensures subprocesses exit: https://github.com/Workiva/dart_dev/pull/318/files#diff-c663bb6e718705f573d16577154614cb

It also supports forwarding the exit signals in the case of detached subprocesses. So far, I haven't hit a scenario where a child process prevents the main process from exiting on the first `^c`, and even if it does, a second `^c` will force it to exit immediately. Does that seem like a sufficient solution to this problem?

greglittlefield-wf:
Yeah that solution seems great! Closing this

PR #298, opened 2019-05-31
Pass serialized config to custom tasks via an environment variable

greglittlefield-wf:
## Motivation
Allow custom tasks to be able to read config without running `dart_dev export-config` (which currently takes a couple seconds due to pub/dart_dev startup time)

## Solution
Somehow pass config to child custom tasks via environment variables

### DIFF
--- a/lib/src/tasks/local/api.dart
+++ b/lib/src/tasks/local/api.dart
@@ -18,10 +18,14 @@ import 'dart:async';
 
 import 'package:dart_dev/util.dart' show TaskProcess;
 
+import 'package:dart_dev/src/tasks/config.dart';
 import 'package:dart_dev/src/tasks/task.dart';
 
 LocalTask local(String executable, Iterable<String> args) {
-  TaskProcess process = new TaskProcess(executable, args);
+  TaskProcess process = new TaskProcess(executable, args, environment: {
+    // Make a serialized version of the config available to this task.
+    'dart_dev_config': serializeConfig(),
+  });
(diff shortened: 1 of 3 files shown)


aviary-wf:
## Security Insights

## Action Items
---

lib/src/tasks/local/api.dart
```diff
@@ -18,10 +18,14 @@ import 'dart:async';
...
 import 'package:dart_dev/util.dart' show TaskProcess;
 
+import 'package:dart_dev/src/tasks/config.dart';
 import 'package:dart_dev/src/tasks/task.dart';
 
 LocalTask local(String executable, Iterable<String> args) {
-  TaskProcess process = new TaskProcess(executable, args);
+  TaskProcess process = new TaskProcess(executable, args, environment: {
+    // Make a serialized version of the config available to this task.
+    'dart_dev_config': serializeConfig(),
```
greglittlefield-wf: Since environment variables shouldn't hold a lot of data, I'm thinking it might be better to write the config to a file, and then make this the path to that file.

evanweible-wf: Yeah I don't see any harm in writing it to something like `.dart_tool/dart_dev/config.json`

evanweible-wf:
Closing this because this pattern doesn't really apply to v3. We still have a use case where sharing config is useful (formatting line length), but I think at this point we're going to try a different approach.

PR #303, opened 2019-06-13
Fix result when test task fails so that correct exit code is returned

smaifullerton-wk:
In the test task, the `outputProcessed` future was only completed if the tests ran or if the test regex didn't match any test names. If anything else happened, the task's `done` future never completed, the result was never returned to `dart_dev_cli.dart`, and no matter what happened in the task, an exit code of `0` was returned.

This PR makes sure that the future is completed in the test task no matter what. As a result, we can remove the extra build that was going on in the test task before. Now, if you run `ddev test` and the build fails, an exit code of 1 will be returned.

### QA +1 Instructions
- [ ] Pull into a repo that builds on Dart 2 that has the `config.test.deleteConflictingOutputs` option set to true.
- [ ] Run `ddev test` with a passing build, ensure that the `--delete-conflicting-outputs` flag is passed to the build command, and tests run successfully.
- [ ] Break something in the repo so the build will fail, re-run `ddev test`. After the build fails, try `echo $?` and ensure that the exit code returned was 0.


aviary-wf:
## Security Insights

## Action Items
---

lib/src/dart_dev_cli.dart
```diff
@@ -182,9 +182,9 @@ Future _run(List<String> args) async {
 
   reporter.log('');
   if (result.successful) {
-    reporter.success(result.message, shout: true);
+    reporter.success(result.message ?? '', shout: true);
```
tomconnell-wf: I'm a little concerned/unclear what empty string will accomplish if result.message is null.

tomconnell-wf:
👏 , but CI failed.

lib/src/dart_dev_cli.dart
smaifullerton-wk: `isNotEmpty` threw an exception if `result.message` is null here: https://github.com/Workiva/dart_dev/blob/master/lib/src/reporter.dart#L102

I'll push a commit to fix this there though. I'm still going to set it to an empty string if null so that it doesn't print "null" in the terminal when there's no `result.message`.

smaifullerton-wk: `isNotEmpty` threw here: https://github.com/Workiva/dart_dev/blob/master/lib/src/reporter.dart#L102 

I will push a commit to fix it up farther down the stack though.

todbachman-wf:
+10

#### Testing Performed
- [x] Pull into `w_router` 
- [x] Set the `config.test.deleteConflictingOutputs` option to `true`
- [x] Run `ddev test` with a passing build
- [x] Confirm  the `--delete-conflicting-outputs` flag is passed to the build command
- [x] Break something in the repo so the build will fail
- [x] Re-run `ddev test`
- [x] Confirm `ddev` returned the exit code `1`
- [x] Fix the failing build
- [x] Re-run `ddev test`
- [x] Confirm `ddev` returned the exit code `0`
- [x] Add `expect(true, isFalse, reason: 'fake news');` to a unit test
- [x] Re-run `ddev test`
- [x] Confirm `ddev` returned the exit code `1`

corwinsheahan-wf:
@Workiva/release-management-p

PR #314, opened 2019-09-13
V3 WIP: Implement analyze, format, sass, test, serve. Add workiva config.

evanweible-wf:


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/tools/format_tool.dart
```diff
@@ -0,0 +1,418 @@
...
+  overwrite,
+}
+
+/// Available dart formatters.
+enum Formatter {
+  // The formatter provided via the Dart SDK.
+  dartfmt,
+  // The formatter provided via the `dart_style` package.
+  dartStyle,
+}
```
maxwellpeterson-wf: Is it worth making both of these an option? Is `dartfmt` not an alias to the other?

lib/configs/workiva.dart
```diff
@@ -0,0 +1,17 @@
+/// An opinionated `tool/dev.dart` base configuration intended to help
+/// standardize dart_dev configuration and command-line usage across Dart
+/// projects in the Workiva organization.
+library dart_dev.configs.workiva;
```
georgelesica-wf: Wasn't this moving to its own package or something?

lib/src/tools/format_tool.dart
evanweible-wf: `dartfmt` does run `dart_style:format`, but with the version of dart_style that is packaged with the Dart SDK. Some packages depend on `dart_style` directly and run that exe, which allows them to pick a specific version or pull in newer versions that are available sooner than they would be via an SDK release.

evanweible-wf: Another way to look at the difference – `dartfmt` works in any dart project, but `pub run dart_style:format` requires `dart_style` be a direct dependency of the current project.

lib/src/tools/analyze_tool.dart
```diff
@@ -0,0 +1,231 @@
...
+  /// as well as any dart files in the project root.
+  List<Glob> include;
+
+  @override
+  FutureOr<int> run([DevToolExecutionContext context]) =>
+      runProcessAndEnsureExit(buildProcess(context ?? DevToolExecutionContext(),
+          configuredAnalyzerArgs: analyzerArgs, include: include));
+
+  @override
+  Command<int> toCommand(String name) => DevToolCommand(name, this,
```
georgelesica-wf: Will this type parameter and the one on line 61 always be the same? If so, maybe the parameter should go on the class?

lib/configs/workiva.dart
evanweible-wf: Yes, I haven't pushed up the commit to the other repo, yet. But when I do, it will be pretty much a direct move from this repo to dart_dev_workiva

lib/src/tools/analyze_tool.dart
evanweible-wf: I was trying to force the usage of an integer exit code via the `DevTool` class – it hardcodes the return type for `FutureOr<int> run()` and `Command<int> toCommand()`. Changing that to a parameterized type would defeat that purpose and allow `DevTool` implementations to return something other than an integer exit code.

lib/src/tools/dart_function_tool.dart
```diff
@@ -0,0 +1,68 @@
...
+
+/// A utility class designed to make it simple to create a [DevTool] from just a
+/// Dart function.
+///
+/// To create a [DartFunctionTool], all that is needed is a function of type:
+///     FutureOr<int> Function(DartFunctionToolContext context)
+///
+/// The `DartFunctionToolContext context` contains info about the context in
+/// which the tool is being run, which can be used in the function. Currently,
+/// it only contains one piece if info: whether or not "verbose" mode is
```
georgelesica-wf: "of"

lib/src/tools/dart_function_tool.dart
```diff
@@ -0,0 +1,68 @@
...
+/// enabled.
+///
+///     import 'package:dart_dev/dart_dev.dart';
+///
+///     int helloWorld(DartFunctionToolContext context) {
+///       print('Hello world!');
+///       return 0;
```
georgelesica-wf: Shell conventions here, right? Should we add a constant for "success" to make it more obvious? Kind of a nit

evanweible-wf: Yeah, there is an `ExitCode` constants class available in `pkg:io` that we could use here.

lib/src/tools/analyze_tool.dart
```diff
@@ -0,0 +1,231 @@
...
+/// This will make it available via the `dart_dev` command-line app like so:
+///     pub run dart_dev analyze
+///
+/// This tool can be configured by modifying any of its fields:
+///     // tool/dev.dart
+///     import 'package:dart_dev/dart_dev.dart';
+///
+///     final config = {
+///       'analyze': AnalyzeTool()
+///         ..dartanalyzerArgs = ['--fatal-infos']
```
smaifullerton-wk: nit: is now `analyzerArgs`

lib/src/tools/format_tool.dart
```diff
@@ -0,0 +1,418 @@
...
+///         ..formatter = Formatter.dartStyle,
+///     };
+///
+/// It is also possible to run this tool directly in a dart script:
+///     FormatTool().run();
+class FormatTool extends DevTool {
+  /// The default mode in which to run the formatter.
+  ///
+  /// This is still overridable via the command line:
+  ///     ddev format -- -n  # dry-run
```
smaifullerton-wk: Can remove the `--`

smaifullerton-wk: can remove the `--`

lib/configs/workiva.dart
evanweible-wf: Moved.

evanweible-wf:
@Workiva/dart-pullassigner

lib/src/tools/dart_function_tool.dart
evanweible-wf: Updated

lib/src/tools/format_tool.dart
evanweible-wf: fixed

lib/src/tools/analyze_tool.dart
evanweible-wf: fixed

README.md
```diff
@@ -1,8 +1,246 @@
...
+$ ddev format
+[INFO] Running subprocess:
+pub run dart_style:format -w <3 paths>
+--------------------------------------
+Unchanged ./lib/foo.dart
+Unchanged ./lib/src/bar.dart
+Formatted ./lib/src/baz.dart
+```
+
+## Extending/Composing Functionality
```
smaifullerton-wk: Are there any other commands that will change the underlying process run based on direct deps? If so, I think we should document them all here, since it's not very intuitive based on how everything else works. First time I ran this I was definitely trying to figure out what `TestTool()` option I had to add to use `build_runner` instead of `pub run test`.

README.md
```diff
@@ -1,8 +1,246 @@
...
+[INFO] Caching finalized dependency graph completed, took 71ms
+[INFO] Creating merged output dir `/var/folders/vb/k8ccjw095q16jrwktw31ctmm0000gn/T/build_runner_testBkm6gS/` completed, took 260ms
+[INFO] Writing asset manifest completed, took 3ms
+[INFO] Succeeded after 12.8s with 1276 outputs (2525 actions)
+Running tests...
+
+00:00 +75: All tests passed!
+```
+
+Additionally, `TestTool` automatically applies [`--build-filter`][build-filter]
```
smaifullerton-wk: What do you think about including the need to specify a test file path for the `--build-filter` to be applied?

README.md
evanweible-wf: Yeah I do want to include more documentation around the individual tools, but I was hoping to keep that info out of the readme and in the api docs for each tool. In general I was just trying to keep the readme focused on the overall idea behind dart_dev (the current readme is pretty huge and tries to document everything but it is kind of a mess because of that).

Mind if I cut a follow up ticket to expand the docs on each of the tools and then include a more prominent reference to those docs in the readme?

smaifullerton-wk: Sounds good

PR #317, opened 2019-10-04
CPLAT-7634: Export a util that builds an input list for the formatter

olesiathoms-wk:
### Problem
The FormatTool has a buildInputs util method that takes a list of exclude and include globs and returns a list of file/dir paths for the dartfmt process. This util will be useful for the over_react_format tool as well, so we should generalize it slightly and export it publicly.

### Acceptance criteria
New util should:
- [ ] List recursively and follow links
- [ ] Not log anything
- [ ] Not catch FileSystemExceptions when listing globs
- [ ] Return an object that contains:
   - [ ] The file paths
   - [ ] The paths that were excluded by an exclude glob
   - [ ] The paths that were skipped because they are Links
   - [ ] The hidden directories (start with a .) that were skipped
- [ ] Be publicly exported from package:dart_dev/utils.dart
For reference, this is the behavior we're trying to match: https://github.com/dart-lang/dart_style/blob/f4d918c4e8e68faee2fac902c868dc368f46f497/lib/src/io.dart#L21

### Testing instructions
- [ ] CI passes
- [ ] Use this version on any repo
  - [ ] [Modify dev.dart](https://github.com/Workiva/dart_dev/blob/3.0.0-wip__tools/README.md#shared-configuration)
  - [ ] Create `symlink` to file (`ln -s source_file [target_file]`)
  - [ ] Create 'symlink' to directory
  - [ ] Set exclude directories
  - [ ] Create hidden directory in one of the subdirectories 
  - [ ] Run `pub run dart_dev format`
  - [ ] Ensure that created links, hidden directories and specified directories to exclude were skipped


aviary-wf:
## Security Insights

## Action Items
---

lib/src/utils/build_inputs.dart
```diff
@@ -0,0 +1,61 @@
...
+/// Builds and returns the list of inputs on which the formatter should be run.
+///
+/// These inputs are determined by expanding the [include] globs and filtering
+/// out any paths that match the expanded [exclude] globs.
+///
+/// Logs may be output in certain scenarios for debugging purposes.
+///
+/// By default these globs are assumed to be relative to the current working
+/// directory, but that can be overridden via [root] for testing purposes.
+Map<String, dynamic> buildInputs(
```
evanweible-wf: Instead of returning a Map, we should create a class to hold the relevant information so that consumers can access it in a type-safe manner rather than relying on particular string keys.

lib/src/utils/build_inputs.dart
```diff
@@ -0,0 +1,61 @@
...
+      'links': <String>[],
+      'excluded': <String>[],
+      'hidden_directories': Set<String>(),
+    },
+  };
+
+  exclude ??= <Glob>[];
+  include ??= <Glob>[];
+  Directory dir = Directory(root ?? '.');
+  for (final entry in dir.listSync(recursive: true, followLinks: false)) {
```
evanweible-wf: Hmm... this got me thinking. Previously, this util would use the list of `include` globs as the starting point for files to format, and then it would filter out any of those files that match one of the `exclude` globs (if there are any).

But, `dartfmt` by default operates on every `.dart` file it can find in the current working directory. Really the only additional functionality we're trying to provide here is a way to exclude certain files, so maybe it would be sufficient to drop the `include` option from the `FormatTool` and only support optional excludes.

In other words, `FormatTool()` would format all dart files by default. If there are certain files that shouldn't be formatted, you'd specify that like so: `FormatTool()..exclude = [Glob('lib/**.g.dart')]`. And that should be sufficient - you'd never need to also tweak the input set since it is already everything.

lib/src/utils/build_inputs.dart
```diff
@@ -0,0 +1,61 @@
...
+        isExcluded = true;
+        continue;
+      }
+    }
+    if (isExcluded) continue;
+
+    if (entry is! File || !entry.path.endsWith('.dart')) continue;
+
+    // If subdirectory starts with ".", ignore it.\
+    if (relative.startsWith('.')) {
```
evanweible-wf: This will only catch paths that have a hidden directory as the first part of the path, e.g. `.dart_tool/build/foo.dart`, but it wouldn't skip something like `example/.pub/foo.dart`.

You can see how the dartfmt tool handles this here: https://github.com/dart-lang/dart_style/blob/f4d918c4e8e68faee2fac902c868dc368f46f497/lib/src/io.dart#L38-L46

lib/src/utils/build_inputs.dart
```diff
@@ -0,0 +1,61 @@
...
+      final hidden_directory = p.split(relative)[0];
+      returnInputs['skipped']['hidden_directories'].add(hidden_directory);
+      continue;
+    }
+
+    if (entry.path.endsWith('.dart')) {
```
evanweible-wf: Line 47 above already handles the `endsWith('.dart')` check, so you can safely remove this if check. The file at this point is guaranteed to be a `.dart` file.

lib/src/utils/build_inputs.dart
evanweible-wf: We should also rename this util slightly since we're making it public. We want it to be clear that it is intended to be used to gather inputs for a dart formatter tool. Maybe `getFormatterInputs`?

lib/src/tools/format_tool.dart
```diff
@@ -273,72 +271,6 @@ FormatExecution buildExecution(
...
-  final excluded = includePaths.intersection(excludePaths);
-  if (excluded.isNotEmpty) {
-    _log.fine('Excluding these paths from formatting:\n  '
-        '${includePaths.intersection(excludePaths).join('\n  ')}');
-  }
-  final inputs = includePaths.difference(excludePaths);
-  if (inputs.isEmpty) {
-    _log.severe('The formatter cannot run because no inputs could be found '
-        'with the configured includes and excludes.\n'
-        'Please modify the excludes and/or includes in "tool/dev.dart".');
```
evanweible-wf: We still want to log the messages that were previously in this function, but we'll need to do it in the `buildExecution` function instead. We should just be able to use the returned object from `getFormatterInputs`

lib/src/utils/formatter_inputs.dart
```diff
@@ -0,0 +1,75 @@
...
+/// - The paths that were skipped because they are links
+/// - The hidden directories(start with a '.') that were skipped
+///
+/// The file paths will include all .dart files in [path] (and its subdirectories),
+/// except any paths that match the expanded [exclude] globs.
+///
+/// By default these globs are assumed to be relative to the current working
+/// directory, but that can be overridden via [root] for testing purposes.
+FormatterInputs getFormatterInputs({List<Glob> exclude, String root}) {
+  FormatterInputs inputs = FormatterInputs(Set(), Set(), Set(), Set());
```
evanweible-wf: It's an anti-pattern to create an object with collection instances and then later mutate those collections. What I was going for with making the fields on the `FormatterInputs` class final was to ensure that the returned object from this function is immutable. I think a better pattern would be to create the `Set`s at the top of this function, and then construct the `FormatterInputs` object at the end when we return, e.g.:

```dart
final included = <String>{};
final excluded = <String>{};
final links = <String>{};
final hidden = <String>{};

...

return FormatterInputs(
  included,
  excludedFiles: excludedFiles,
  hiddenDirectories: hiddenDirectories,
  skippedLinks: skippedLinks);
```

(Also, the dart style guide now recommends using the collection literals when possible: https://dart.dev/guides/language/effective-dart/usage#do-use-collection-literals-when-possible)

lib/src/utils/formatter_inputs.dart
```diff
@@ -0,0 +1,75 @@
...
+
+  exclude ??= <Glob>[];
+
+  if (exclude.isEmpty) inputs.includedFiles.add(root ?? '.');
```
evanweible-wf: If there are no excludes, we should bail out early and return just the directory. Also, I told you wrong earlier when you asked about this, in the case where `root` is not null, we should still return `.` – that is because the executable will be run from that same `root` path, so we want the paths to be relative to that dir.
```dart
if (exclude.isEmpty) {
  return FormatterInputs({'.'});
}
```

lib/src/utils/formatter_inputs.dart
```diff
@@ -0,0 +1,75 @@
...
+      }
+    }
+  }
+
+  return inputs;
+}
+
+class FormatterInputs {
+  FormatterInputs(this.includedFiles, this.links, this.excludedFiles,
+      this.hiddenDirectories);
```
evanweible-wf: In the case where there are no exclude globs, we want to just return the root directory, which means there would be no skipped links, hidden directories, nor excluded files. To better support that, we could make the last three params optional.

lib/src/utils/formatter_inputs.dart
```diff
@@ -0,0 +1,75 @@
...
+
+  final Set<String> includedFiles;
+
+  final Set<String> links;
```
evanweible-wf: I think `skippedLinks` would be more explicit in its meaning.

test/tools/format_tool_test.dart
```diff
@@ -172,8 +208,13 @@ void main() {
...
         expect(execution.exitCode, isNull);
         expect(execution.process.executable, 'pub');
-        expect(execution.process.args,
-            orderedEquals(['run', 'dart_style:format', '.']));
+        expect(
+            execution.process.args,
+            orderedEquals([
+              'run',
+              'dart_style:format',
+              'test/tools/fixtures/format/has_dart_style'
```
evanweible-wf: In this case, we want the returned path to be relative to the path we passed to `buildExecution`. So I would expect this to still be `.` and not the full path to the test fixture

lib/src/tools/format_tool.dart
```diff
@@ -85,8 +85,7 @@ class FormatTool extends DevTool {
         configuredFormatterArgs: formatterArgs,
```
evanweible-wf: We should remove the `include` config option above on line 79

evanweible-wf:
QA +1
- [x] CI passes

PR #338, opened 2020-02-28
Add Noop DevTool

laurameckley-wk:
In DPC, we've had quite a few instances where people try to run tests in the root directory, and it appears to try to run tests despite not having a testing command defined in the config. I propose we add a NoopTool to enable us to define any command as a noop. For us, we could leverage it by assigning 'test' in root to the NoopTool, giving a concrete error message.

### DIFF
--- a/lib/dart_dev.dart
+++ b/lib/dart_dev.dart
@@ -8,5 +8,6 @@ export 'src/tools/format_tool.dart'
     show FormatMode, Formatter, FormatterInputs, FormatTool;
 export 'src/tools/process_tool.dart' show ProcessTool;
 export 'src/tools/test_tool.dart' show TestTool;
+export 'src/tools/noop_tool.dart' show NoopTool;
 export 'src/tools/tuneup_check_tool.dart' show TuneupCheckTool;
 export 'src/tools/webdev_serve_tool.dart' show WebdevServeTool;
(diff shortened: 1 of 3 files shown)


aviary-wf:
## Security Insights

## Action Items
---

evanweible-wf:
@laurameckley-wk I think you could accomplish this with `DevTool.fromFunction()` – would this work for your use case?

```dart
// tool/dart_dev/config.dart
import 'package:io/ansi.dart';
import 'package:io/io.dart';

final config = {
  'test': DevTool.fromFunction((_) {
    print(red.wrap('The "test" command is not supported from the root.'));
    return ExitCode.usage.code;
  }),
};
```

evanweible-wf:
And if you have multiple commands that need that same treatment, you could do something like this:

```dart
// tool/dart_dev/config.dart
import 'package:io/ansi.dart';
import 'package:io/io.dart';

final config = {
  ...unsupportedCommands(['test', 'foo']),
};

Map<String, DevTool> unsupportedCommands(Iterable<String> names) => {
      for (final name in names)
        name: DevTool.fromFunction((_) {
          print(
              red.wrap('The "$name" command is not supported from the root.'));
          return ExitCode.usage.code;
        })
    };

```

laurameckley-wk:
@evanweible-wf That works perfectly, thanks for the recommendation!

PR #340, opened 2020-04-15
CPLAT-10490 - Have format_tool add whole directories 💼

tomconnell-wf:

PR #341, opened 2020-04-15
CPLAT-10490 - Add a config that allows collapsing directories with no excludes. 🏋️‍♀️

tomconnell-wf:
# Problem
We have a repository that has over 3000 files to format.  Mac OS, at least, had a hardcoded kernel setting that causes that argument list to be too long.  Because of that, we aren't currently able to format in one pass.

# Solution
Allow the format tool to collapse directories that have no excludes.  In our repo, this cut the paths from 3012 to 705.  The default still expands directories, so consumers will not experience a change in behavior unless they opt-in.

# Testing
- [ ] Pull this version of dart_dev into your repo.  Run dart_dev format, and verify it formats the same files as before, but using many less input paths.

### Without Collapse

```
$ doc_plat_client git:(PP-3396_blackline_review_cards_rebase) ✗ time ddev format
Matching exclude glob dirs:
[INFO] excluding 728 files
[INFO] Running subprocess:
pub run over_react_format --line-length=120 <2835 paths>
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Formatting 2835 files...
    2835 files unchanged
Success! All files are already formatted correctly.
pub run dart_dev format  79.78s user 11.79s system 334% cpu 27.380 total
```

### With Collapse

```
$doc_plat_client git:(PP-3396_blackline_review_cards_rebase) ✗ time ddev format
Matching exclude glob dirs:
Glob didn't match any paths
[INFO] excluding 728 files
[INFO] Running subprocess:
pub run over_react_format --line-length=120 <707 paths>
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Formatting 2835 files...
    2835 files unchanged
Success! All files are already formatted correctly.
pub run dart_dev format  77.91s user 11.95s system 342% cpu 26.256 total
```


tomconnell-wf:
Rebased from https://github.com/Workiva/dart_dev/pull/340

aviary-wf:
## Security Insights

## Action Items
---

tomconnell-wf:
I'll need to make a tiny PR to over_react_format to have it pass through a collapseDirectories option.

tomconnell-wf:
With things provisionally wired up through over_react_format:
image

Notice the number of paths passed in versus the number of files formatted.  They used to be lockstep.

lib/src/tools/format_tool.dart
```diff
@@ -161,20 +198,49 @@ class FormatTool extends DevTool {
...
       } else {
-        if (entry is File) includedFiles.add(relative);
+        if (collapseDirectories && entry is Directory) {
+          _log.finest('directory: $entry');
+          currentDirectory = relative;
+          // It seems we can rely on the order of files coming from listSync.
+          // If a directory does not match any of the globs, and does not contain any of the globs,
+          // we should be able to just add that directory and skip adding any of its children files or directories.
+          if (exclude.any((glob) => p.isWithin(
+              entry.path, p.relative(glob.toString(), from: dir.path)))) {
```
evanweible-wf, line 220: I don't think this works. For one, it's treating `glob.toString()` as a filepath, when it's actually a pattern. For example, it's common to exclude generated files with a `Glob('**.g.dart')` glob which matches any file in the project with a `.g.dart` extension. But used here as a filepath, you'd get false negatives:

```dart
p.isWithin('lib/', '**.g.dart') // false
```

Another example would be using `{}` to list multiple options:

```dart
p.isWithin('lib/src/', 'lib/{src,internal}/*.dart') // false
```

In both of these cases, we should be matching but aren't.

---

To fully support this functionality, I think we'd need to expand the list of files and then go back and test them against the exclude globs using `Glob.matches()`, collapsing directories one at a time if none of its children match.

evanweible-wf: This implementation happens to work in DPC because the excludes are only using patterns to target whole directories (e.g. `app/build/**`), and the rest are actual file paths.

tomconnell-wf:
@evanweible-wf I think this addresses the problem.  It uses Glob.listSync, which gets all the FileSystemEntities to be excluded.  I think make a set of the directories that contain those files.  Then, when going through the files to format, if a directory contains any of the excluded directory set, it keeps descending and does not collapse.  The default behavior should remain the same as before this PR.

tomconnell-wf:
I tested on a wider array of repos that I knew were using dart_dev ^3.0.0 and saw the same files formatted with and without collapse.

tomconnell-wf:
@smaifullerton-wk fyi

lib/src/tools/format_tool.dart
```diff
@@ -58,6 +58,11 @@ class FormatTool extends DevTool {
   /// By default, nothing is excluded.
   List<Glob> exclude;
 
+  /// If true, pass whole directories with no exclusions to the formatter.
+  ///
+  /// This is helpful for large projects that may cause the formatter command to have too many arguments
+  bool collapseDirectories = false;
```
evanweible-wf, line 64: With the updated implementation it seems like we could just do this by default and avoid putting this behind a flag at all.

tomconnell-wf: I will test consumers tomorrow.

tomconnell-wf:
I tested doc_plat_client, graph_app, wdesk_sdk, audit, w_outline, app_intelligence, and over react with collapse behavior and without.  It wound up excluding the same number of files and formatting the same number of files.  I checked the actual list of excluded files in a couple of the repos, and they were the same.

Of interest was w_comments.  It has no excludes, so doesn't benefit from the collapse behavior, but since there isn't a performance implication, I don't want to look into it.

evanweible-wf:
Looks awesome, thanks for doing this! I'll get this merged and release soon.

olesiathoms-wk:
@Workiva/release-management-p

PR #344, opened 2020-07-13
CPLAT-11875: Add Fast Format Command

joebingham-wk:
# [CPLAT-11875](https://jira.atl.workiva.net/browse/CPLAT-11875)
Issue Status

Added a `hackFastFormat` command to bypass executable generation and just run the formatting command.

## QA'ing
Generally, to run the command globally enable / run the package and run `pub global run dart_dev hackFastFormat {path(s)}`.

Steps:
- Following readme instructions to install the file watcher
- Make sure to globally enable this branch locally
- Update the watcher to execute `pub global run ...` (just need to add the "global" key word)
- Run through the scenarios below
  - Note that it may be beneficial to run the command via cli just to verify the logged command is as expected

Possible scenarios to QA:
- [ ] When the project uses the ORF tool and:
  - [ ] does / doesn't specify a line-length
- [ ] When the project uses the Format tool and:
  - [ ] the project uses the `dartStyle` formatter
  - [ ] the project uses the `dartfmt` formatter
  - [ ] does / doesn't specify command args
- [ ] When either tool specifies a specific file
- [ ] When either tool does not specify a specific file (should fail)


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/utils/config_visitor.dart
```diff
@@ -0,0 +1,78 @@
...
+      }
+
+      return false;
+    }, orElse: () => null);
+
+    if (formatKey != null) {
+      if (node.value.runtimeType.toString() != 'FormatTool') {
+        usesCustomFormatter = true;
+
+        if (node.value.runtimeType.toString() != 'OverReactFormatTool') {
```
joebingham-wk, line 43: Note to self - this is off and should be (assuming the core logic is good):
```suggestion
        if (node.value.runtimeType.toString() == 'OverReactFormatTool') {
```

joebingham-wk:
Note: I added a compiled executable for the tool for testing / demo purposes to imitate what it will be like after the pre-compilation takes place. It should be removed before merge!

lib/src/executable.dart
```diff
@@ -57,9 +64,53 @@ Future<void> run(List<String> args) async {
   exitCode = await process.exitCode;
 }
 
+Future<void> handleFastFormat(List<String> args) async {
```
evanweible-wf, line 67: So the key here in terms of speed is to reconstruct the formatting tool (whichever one it is) from `tool/dart_dev/config.dart` without having to actually load that dart file, right? Because then it can be completely precompiled?

I think we could go a bit further for the non-ORF use case and reconstruct two config options: the formatter to use (dart_style vs dartfmt) and the command line args. If we do that, then all three formatters along with custom line length should be supported.

One idea for the implementation here would be to modify the visitor to reconstruct the format tool (either OverReactFormatTool, FormatTool, or nothing if no `format` entry is found), e.g.:
```dart
Future<void> handleFastFormat(List<String> args) async {
  assertDirIsDartPackage();

  DevTool formatTool;
  final configFile = File(_configPath);
  if (configFile.existsSync()) {
    final configVisitor = ConfigVisitor();
    parseString(content: configFile.readAsStringSync())
        .unit
        .accept(configVisitor);
    formatTool = configVisitor.formatTool; // could be null if no custom `format` entry found
  }
  formatTool ??= chooseDefaultFormatTool();

  try {
    exitCode = await DartDevRunner({'hackFastFormat': formatTool}).run(args);
  } ...
}

DevTool chooseDefaultFormatTool() => ...; // https://github.com/Workiva/dart_dev_workiva/blob/master/lib/src/workiva_config.dart#L34-L46  
```

lib/src/utils/format_tool_builder.dart
```diff
@@ -0,0 +1,234 @@
...
+    if (mapEntryKey != 'format') return;
+
+    final formatterInvocation = node.value;
+    formatDevTool = detectFormatter(node.value);
+
+    if (formatterInvocation is CascadeExpression) {
+      List<Glob> reconstructedExcludesList = [];
+
+      final excludeExpression =
+          formatterInvocation.getCascadeByProperty('exclude');
```
evanweible-wf, line 49: Do we need to reconstruct the excludes if we're using the rest args as the inputs instead of finding inputs like we normally would?

lib/src/utils/format_tool_builder.dart
```diff
@@ -0,0 +1,234 @@
...
+    detectedFormatterName =
+        formatterNode.target.toSource().replaceAll(RegExp('[()]'), '');
+  }
+
+  if (detectedFormatterName == 'FormatTool') {
+    tool = FormatTool();
+  } else if (detectedFormatterName == 'OverReactFormatTool') {
+    tool = OverReactFormatTool();
+  } else {
+    // FIXME handle unknown formatter
```
evanweible-wf, line 220: Still need to do this

lib/src/utils/format_tool_builder.dart
joebingham-wk: We do not, I just thought while I was here I could add it so that `hackFastFormat` command works correctly without using the rest args in case a need ever existed. If it's detrimental to the goal of the ticket, I can remove it though!

lib/src/utils/format_tool_builder.dart
```diff
@@ -0,0 +1,233 @@
...
+  @override
+  visitMapLiteralEntry(MapLiteralEntry node) {
+    super.visitMapLiteralEntry(node);
+
+    String mapEntryKey = node.key.toSource().replaceAll("'", "");
+
+    if (mapEntryKey != 'format') return;
+
+    final formatterInvocation = node.value;
+    formatDevTool = detectFormatter(node.value);
```
evanweible-wf, line 39: [nit]
```suggestion
    formatDevTool = detectFormatter(formatterInvocation);
```

lib/src/utils/format_tool_builder.dart
```diff
@@ -0,0 +1,233 @@
...
+          reconstructedExcludesList = detectedExcludesList.elements
+              .whereType<MethodInvocation>()
+              .where((e) {
+            return e.argumentList.arguments.length == 1 &&
+                e.argumentList.arguments.first is StringLiteral;
+          }).map((e) {
+            final path =
+                (e.argumentList.arguments.first as StringLiteral).stringValue;
+            return Glob(path);
+          }).toList();
```
evanweible-wf, line 63: I think this would miss excluded globs that use any of the [optional constructor params](https://pub.dev/documentation/glob/latest/glob/Glob/Glob.html). Maybe instead of filtering based on method invocations that only have 1 String arg, we could filter by the `Glob()` constructor name and only use the first arg?

lib/src/utils/format_tool_builder.dart
evanweible-wf: I'm still a little hesitant to even add this support for reconstructing excludes when we don't currently have a use case for it, especially given the code below and the possibility that warnings might be emitted that won't actually cause any issues in practice.

lib/src/utils/format_tool_builder.dart
```diff
@@ -0,0 +1,233 @@
...
+        final lineLengthAssignment =
+            formatterInvocation.getCascadeByProperty('lineLength');
+        if (reconstructedExcludesList.isNotEmpty) {
+          typedFormatDevTool.exclude = reconstructedExcludesList;
+        }
+        if (lineLengthAssignment != null) {
+          final lengthExpression = lineLengthAssignment.rightHandSide;
+          if (lengthExpression is IntegerLiteral) {
+            (formatDevTool as OverReactFormatTool).lineLength =
+                lengthExpression.value;
```
evanweible-wf, line 122: [nit]
```suggestion
            typedFormatDevTool.lineLength = lengthExpression.value;
```

lib/src/tools/over_react_format_tool.dart
```diff
@@ -0,0 +1,59 @@
...
+    String executable, Iterable<String> inputs, Iterable<String> args,
+    {bool verbose}) {
+  verbose ??= false;
+  final exeAndArgs = '$executable ${args.join(' ')}'.trim();
+  if (inputs.length <= 5 || verbose) {
+    logSubprocessHeader(log, '$exeAndArgs ${inputs.join(' ')}');
+  } else {
+    logSubprocessHeader(log, '$exeAndArgs <${inputs.length} paths>');
+  }
+}
```
evanweible-wf, line 59: This is identical to https://github.com/Workiva/dart_dev/blob/3860a1bd1816ecc22ba7059675e8e969565e0365/lib/src/tools/format_tool.dart#L479-L494 - should just be able to use that instead.

lib/src/utils/format_tool_builder.dart
```diff
@@ -0,0 +1,233 @@
...
+
+extension on CascadeExpression {
+  AssignmentExpression getCascadeByProperty(String property) {
+    return cascadeSections.whereType<AssignmentExpression>().firstWhere(
+        (assignment) {
+      final lhs = assignment.leftHandSide;
+      return lhs is PropertyAccess && lhs.propertyName.name == property;
+    }, orElse: () => null);
+  }
+}
```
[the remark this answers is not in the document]
joebingham-wk: Yep, that's definitely fair!

lib/src/executable.dart
```diff
@@ -30,25 +36,34 @@ final _relativeDevDartPath = p.relative(
...
   }
 
+  if (args.contains('hackFastFormat') && !oldDevDartExists) {
+    await handleFastFormat(args);
+
+    if (exitCode == ExitCode.config.code) {
+      log.severe('Failed to reconstruct the format tool\'s configuration.\n\n'
+          'This is likely because dart_dev expects either the FormatTool class or the\n'
+          'OverReactFormatTool class.');
+    }
```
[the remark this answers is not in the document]
joebingham-wk: Good catch!

lib/src/executable.dart
```diff
@@ -57,9 +72,41 @@ Future<void> run(List<String> args) async {
...
+  if (configFile.existsSync()) {
+    final toolBuilder = FormatToolBuilder();
+    parseString(content: configFile.readAsStringSync())
+        .unit
+        .accept(toolBuilder);
+    formatTool = toolBuilder
+        .formatDevTool; // could be null if no custom `format` entry found
+
+    if (formatTool == null && toolBuilder.failedToDetectAKnownFormatter) {
+      exitCode = ExitCode.config.code;
```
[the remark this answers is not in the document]
joebingham-wk: Yep, that's a good idea!

lib/src/utils/format_tool_builder.dart
joebingham-wk: 👍  No problem! You're definitely right that it could cause issues and a more fully-baked implementation would only do this when necessary and not all the time. So I'll save the work as a spike in case it ever becomes important but can remove it from this PR!

lib/src/utils/format_tool_builder.dart
joebingham-wk: Just removed since we don't support it!

README.md
```diff
@@ -231,8 +232,57 @@ final config = {
 };
 ```
 
+## Format on save
+dart_dev can be used to facilate formatting on save inside of JetBrains IDEs. For setup instructions, see below.
+
+__NOTE:__ This functionality is reliant upon the `pub run dart_dev` process running at least version `3.6.0` of this tool. Ensure that your `pubspec.yaml` specifies an appropriate version range, otherwise the method described below will not work.
```
joebingham-wk, line 238: Similar to the ORF change a while back, this min version assumes we'll bump ddev a minor for adding a new API

lib/src/tools/over_react_format_tool.dart
```diff
@@ -0,0 +1,49 @@
...
+            'Running the command over an entire project may format files that '
+            'would be excluded using the standard "format" command.');
+      } else {
+        logError(
+            '"hackFastFormat" must specify targets to format.\n',
+            'hackFastFormat should only be used to format specific files. '
+                'Running the command over an entire project may format files that '
+                'would be excluded using the standard "format" command.');
+        return ExitCode.usage.code;
+      }
```
evanweible-wf, line 37: `context` shouldn't ever be null when actually running via `pub run dart_dev`, so is this branch just for unit tests? If that's the case, I'd say we get rid of this (and `logError()`) and just make sure we pass in a `DevToolExecutionContext` argument when we want to test the above branch.

README.md
```diff
@@ -231,8 +232,57 @@ final config = {
...
+
+### JetBrains IDEs (WebStorm, IntelliJ, etc.)
+Webstorm exposes a File Watcher utility that can be used to run commands when a file saves. For this approach, all you need to do is set up the file watcher. Shoutout to @patkujawa-wf for creating the original inspiration of this solution!
+
+#### `hackFastFormat`
+To facilitate the formatting on save capability, the command `hackFastFormat` was introduced. It is important because the standard format command is not optimized for this situation. `hackFastFormat` replicates the core behavior necessary for running the formatter, but excludes some functionality that the standard format command has. Consequently, it may behave differently. While it can be run directly from a shell, that is not the intended usecase.
+
+##### Known Limitations
+- Can only recognize the `FormatTool` and `OverReactFormatTool`. If you use a custom formatter, this will not work for you.
+- Can only run on targeted files. Because this command is optimized for formatting on save, the assumption is that it will only run on a single, specific file. If run without a target file, the command will throw a usage error.
```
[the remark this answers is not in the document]
joebingham-wk: That's a really good point. I thought the first point was still worth mentioning to consumers are aware, but I re-organized it a little to make more sense!

lib/src/tools/over_react_format_tool.dart
joebingham-wk: 👍  Sweeet, that makes things easier. I originally had a fallback because `run` only takes in context as an optional parameter, so I didn't want to not include a case for that. If we're comfortable with it though, it's much nicer to assume it will be there

evanweible-wf:
LGTM! I can test out the VS Code side of things if someone else can test out the JetBrains setup.

joebingham-wk:
@sydneyjodon-wk will QA the file watcher when she gets a chance!

sydneyjodon-wk:
QA+1 using WebStorm

- [X] using `FormatTool` with `dartfmt` formatter
  - [X] tested with and without line-length arg
  - [X] tested on whole project, specific folder, and specific file 
  - [X] fails when file not specified
- [X] using `FormatTool` with `dartStyle` formatter
  - [X] tested with and without line-length arg
  - [X] tested on whole project, specific folder, and specific file 
  - [X] fails when file not specified
- [X] using `OverReactFormatTool`
  - [X] tested with and without line-length
  - [X] tested on whole project, specific folder, and specific file 
  - [X] fails when file not specified
- [X] tested with no format tool set - infers formatter

lib/src/tools/over_react_format_tool.dart
```diff
@@ -21,20 +20,11 @@ class OverReactFormatTool extends DevTool {
...
-            'would be excluded using the standard "format" command.');
-      } else {
-        logError(
-            '"hackFastFormat" must specify targets to format.\n',
-            'hackFastFormat should only be used to format specific files. '
-                'Running the command over an entire project may format files that '
-                'would be excluded using the standard "format" command.');
-        return ExitCode.usage.code;
-      }
+      context.usageException(
```
evanweible-wf, line 23: Sorry, I missed this when you addressed my previous comment. You're right that `context` is optional here, so we should still be protecting against a `null` value:
```suggestion
      context?.usageException(
```

I was just explaining that you don't need to implement a separate logging mechanism for when it is null because the only time that can happen is during unit tests or when someone instantiates this tool and runs it manually (which won't happen since it isn't public).

joebingham-wk: Ohh yep, that makes sense haha. Awesome, updated that!!

evanweible-wf:
QA +1 refresh, last two commits shouldn't change anything from: https://github.com/Workiva/dart_dev/pull/344#issuecomment-672322656

evanweible-wf:
@Workiva/release-management-p

joebingham-wk:
+10

colefeisthamel-wf:
RM +1 team has provided reviews on this PR

PR #354, opened 2021-09-02
CPLAT-15446: Improve criteria for using `build_runner` for tests

evanweible-wf:
Now only uses `build_runner` to run tests if the package has a direct dependency on both `build_runner` and `build_test` (previously we only checked for `build_test`). For packages that contain builder implementations, they will likely have a dependency on `build_test` for use in their builder tests, but don't need to run tests via `build_runner`.

### DIFF
--- a/lib/src/tools/test_tool.dart
+++ b/lib/src/tools/test_tool.dart
@@ -162,12 +162,12 @@ class TestExecution {
 /// Builds and returns the full list of args for the test process that
 /// [TestTool] will start.
 ///
-/// If [useBuildTest] is true, the returned args will run tests via
+/// If [useBuildRunner] is true, the returned args will run tests via
 /// `pub run build_runner test`. Additional args targeting the build process
 /// will immediately follow and args targeting the test process will follow an
 /// arg separator (`--`).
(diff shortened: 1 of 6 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/tools/test_tool.dart
```diff
@@ -186,10 +186,10 @@ List<String> buildArgs({
   ArgResults argResults,
   List<String> configuredBuildArgs,
   List<String> configuredTestArgs,
-  bool useBuildTest,
```
evanweible-wf, line 189: fyi only the `TestTool` class in this file is publicly exported, so this param name change is safe.

corwinsheahan-wf (approved):

todbachman-wf (approved):

todbachman-wf:
QA +1. CI passed and it looks like you've unit tested all the scenarios correctly. 🎉

evanweible-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #362, opened 2021-10-12
CPLAT-15507: ddev should use new dart CLI

annawatson-wk:
Purpose:

ddev analyze runs dartanalyzer, which does not leverage the analysis server
ddev format may run dartfmt (unless the package depends on dart_style)
in several places, ddev will run other dart executables via pub run
As of Dart 2.10 (https://medium.com/dartlang/announcing-dart-2-10-350823952bd5), there exists a new unified dart CLI with improved options for analysis, formatting, and running packages. The primary benefit to this new CLI is dart analyze which (like tuneup) uses the analysis server, which is important because the now-outdated dartanalyzer suffers from several deficiences and bugs because it doesn't use the analysis server.

Note: dart_dev also provides a tuneup-based option for analysis. There does still seem to be some functionality in tuneup that doesn't exist in dart analyze, but its future is unclear. We can probably leave the tuneup tool in place, but recommend that the default AnalyzeTool be used once it is updated to use dart analyze.

Acceptance Criteria
- Update AnalyzeTool to have the option to run dart analyze instead of dartanalyzer
- Update FormatTool to have the option run dart format instead of dartfmt (this shouldn't change the scenario when dart_style is used)
- Update any usage of pub run to dart run/ except dart test
- Update any usage of pub global to dart pub global
- Update any usage of pub <get|upgrade> to dart pub <get|upgrade>
- Update any applicable references to outdated CLI commands in the README and documentation guides

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,10 +110,9 @@ memorized or referenced in order to run said task.
 Consider formatting as an example. The default approach to formatting files is
 to run `dartfmt -w .`. But, some projects may want to exclude certain files that
 would otherwise be formatted by this command. Or, some projects may want to use
-`dart run dart_style:format` instead of `dartfmt` or `dart format` instead of 
-`dartfmt`. Currently, there is no project-level configuration supported by the 
-formatter, so these sorts of things just have to be documented in a `README.md` 
-or `CONTRIBUTING.md`.
+`pub run dart_style:format` instead of `dart format`. Currently, there is no
+project-level configuration supported by the formatter, so these sorts of things
+just have to be documented in a `README.md` or `CONTRIBUTING.md`.
 
 With `dart_dev`, this can be accomplished like so:
 
@@ -145,10 +144,10 @@ Using existing tooling provided by (or conventionalized by) the Dart community
 should always be the goal, but the reality is that there are gaps. Certain use
 cases can be made more convenient and new use cases may arise.
 
-Consider test running as an example. For simple projects, `dart run test` is
+Consider test running as an example. For simple projects, `dart test` is
 sufficient. In fact, the test package supports a huge amount of project-level
 configuration via `dart_test.yaml`, which means that for projects that are
-properly configured, `dart run test` just works.
+properly configured, `dart test` just works.
 
 Unfortunately, at this time, projects that rely on builders must run tests via
 `dart run build_runner test`. Based on the project, you would need to know which
@@ -157,13 +156,13 @@ test command should be run.
 With `dart_dev`, the `TestTool` handles this automatically by checking the
 project's `pubspec.yaml` for a dependency on `build_test`. If present, tests
 will be run via `dart run build_runner test`, otherwise it falls back to the
-default of `dart run test`.
+default of `dart test`.
 
 ```bash
 # In a project without a `build_test` dependency:
 $ ddev test
 [INFO] Running subprocess:
-dart run test
+dart test
 ----------------------------
 00:01 +75: All tests passed!
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ By default, this provides three core tasks:
 Run any of these tools via the `dart_dev` command-line app:
 
 ```bash
-$ pub run dart_dev analyze
+$ dart run dart_dev analyze
 [INFO] Running subprocess:
 dartanalyzer .
 --------------------------
@@ -51,7 +51,7 @@ No issues found!
 > We recommend adding a `ddev` alias:
 >
 > ```bash
-> alias ddev='pub run dart_dev'
+> alias ddev='dart run dart_dev'
 > ```
 
 Additional Dart developer tools can be added and every tool can be configured.
@@ -79,7 +79,7 @@ final config = {
 Most Dart projects eventually share a common set of development requirements
 (e.g. static analysis, formatting, test running, serving, etc.). The Dart SDK
 along with some core packages supply the necessary tooling for these developer
-tasks (e.g. `dartanalyzer`, `dartfmt`, or `pub run test`).
+tasks (e.g. `dartanalyzer`, `dartfmt`, or `dart run test`).
 
 While the core tooling gets us far, there are two areas in which we feel it
 falls short:
@@ -110,7 +110,7 @@ memorized or referenced in order to run said task.
 Consider formatting as an example. The default approach to formatting files is
 to run `dartfmt -w .`. But, some projects may want to exclude certain files that
 would otherwise be formatted by this command. Or, some projects may want to use
-`pub run dart_style:format` instead of `dartfmt`. Currently, there is no
+`dart run dart_style:format` instead of `dartfmt`. Currently, there is no
 project-level configuration supported by the formatter, so these sorts of things
 just have to be documented in a `README.md` or `CONTRIBUTING.md`.
 
@@ -131,7 +131,7 @@ final config = {
 ```bash
 $ ddev format
 [INFO] Running subprocess:
-pub run dart_style:format -w <3 paths>
+dart run dart_style:format -w <3 paths>
 --------------------------------------
 Unchanged ./lib/foo.dart
 Unchanged ./lib/src/bar.dart
@@ -144,25 +144,25 @@ Using existing tooling provided by (or conventionalized by) the Dart community
 should always be the goal, but the reality is that there are gaps. Certain use
 cases can be made more convenient and new use cases may arise.
 
-Consider test running as an example. For simple projects, `pub run test` is
+Consider test running as an example. For simple projects, `dart run test` is
 sufficient. In fact, the test package supports a huge amount of project-level
 configuration via `dart_test.yaml`, which means that for projects that are
-properly configured, `pub run test` just works.
+properly configured, `dart run test` just works.
 
 Unfortunately, at this time, projects that rely on builders must run tests via
-`pub run build_runner test`. Based on the project, you would need to know which
+`dart run build_runner test`. Based on the project, you would need to know which
 test command should be run.
 
 With `dart_dev`, the `TestTool` handles this automatically by checking the
 project's `pubspec.yaml` for a dependency on `build_test`. If present, tests
-will be run via `pub run build_runner test`, otherwise it falls back to the
-default of `pub run test`.
+will be run via `dart run build_runner test`, otherwise it falls back to the
+default of `dart run test`.
 
 ```bash
 # In a project without a `build_test` dependency:
 $ ddev test
 [INFO] Running subprocess:
-pub run test
+dart run test
 ----------------------------
 00:01 +75: All tests passed!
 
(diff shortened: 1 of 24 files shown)


lib/src/tools/analyze_tool.dart
```diff
@@ -52,25 +53,33 @@ class AnalyzeTool extends DevTool {
   /// files in the current working directory.
   List<Glob> include;
 
+  /// The default tool for analysis will be `dartanalyzer` unless opted in here
+  /// to utilize `dart analyze`.
```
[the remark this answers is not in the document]
evanweible-wf: I think we can avoid needing to document this explicitly (other than in the changelog) since the package has its SDK minimum set to a version that safely ensures these commands are available:
https://github.com/Workiva/dart_dev/blob/master/pubspec.yaml#L7

lib/src/tools/format_tool.dart
```diff
@@ -366,6 +373,42 @@ Iterable<String> buildArgs(
...
+    Iterable<String> executableArgs, FormatMode mode,
+    {ArgResults argResults, List<String> configuredFormatterArgs}) {
+  final args = <String>[
+    ...executableArgs,
+
+    // Combine all args that should be passed through to the dart format in this
+    // order:
+    // 1. Mode flag(s), if configured
+    if (mode == FormatMode.check) '--set-exit-if-changed',
+    if (mode == FormatMode.dryRun) ...['-o', 'none', '.'],
```
evanweible-wf, line 402: The `.` at the end shouldn't be included here since we add the files/dirs to format below on line 525
```suggestion
    if (mode == FormatMode.dryRun) ...['-o', 'none'],
```

test/tools/analyze_tool_test.dart
```diff
@@ -26,35 +26,60 @@ void main() {
 
   group('buildArgs', () {
     test('defaults to an empty list', () {
-      expect(buildArgs(), isEmpty);
+      expect(buildArgs(useDartAnalyze: false), isEmpty);
```
evanweible-wf, line 29: With the null-initialization changes for this param mentioned above, you should be able to revert this change and have the test pass again (same with the instances below where `useDartAnalyze: false` was added)

lib/src/tools/format_tool.dart
```diff
@@ -464,26 +508,36 @@ FormatExecution buildExecution(
...
+  final dartFormatter = buildProcess(formatter);
+  Iterable<String> args;
+  if (formatter == Formatter.dartFormat) {
+    args = buildArgsForDartFormat(dartFormatter.args, mode,
+        argResults: context.argResults,
+        configuredFormatterArgs: configuredFormatterArgs);
+  } else {
+    args = buildArgs(dartFormatter.args, mode,
+        argResults: context.argResults,
+        configuredFormatterArgs: configuredFormatterArgs);
```
annawatson-wk, line 520: Thoughts on utilizing `buildArgs` for both `dartfmt` & `dart format`? I assume eventually we're going to want to eliminate the use of `dartfmt` so I thought seperating the logic would make that transition easier/ the unit tests are also seperated.

lib/src/tools/format_tool.dart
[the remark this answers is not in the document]
evanweible-wf: Oh good callout, I missed this the first time through. The only thing missing here is that we want check mode to essentially be a superset of dry-run mode, meaning that it should look like so:
```
    if (mode == FormatMode.check) ...['-o', 'none', '--set-exit-if-changed'],
    if (mode == FormatMode.dryRun) ...['-o', 'none'],
```

lib/src/tools/format_tool.dart
evanweible-wf: Yeah I think this is a good compromise. Also, it looks like [even `dart_style:format` will eventually move to the same CLI](https://github.com/dart-lang/dart_style/issues/986) that `dart format` has now, so eventually we can use that new logic for both.

lib/src/tools/analyze_tool.dart
```diff
@@ -175,9 +193,13 @@ void logCommand(
   Iterable<String> args,
   Iterable<String> entrypoints, {
   bool verbose,
+  bool useDartAnalyzer,
 }) {
```
evanweible-wf, line 197: alphabetize these two params

test/tools/analyze_tool_test.dart
```diff
@@ -163,19 +206,27 @@ void main() {
...
+      logCommand(['-t'], ['a', 'b', 'c', 'd', 'e', 'f'],
+          useDartAnalyzer: false);
     });
 
     test('with >5 entrypoints in verbose mode', () {
       expect(Logger.root.onRecord,
           emitsThrough(infoLogOf(contains('dartanalyzer -t a b c d e f'))));
-      logCommand(['-t'], ['a', 'b', 'c', 'd', 'e', 'f'], verbose: true);
+      logCommand(['-t'], ['a', 'b', 'c', 'd', 'e', 'f'],
+          verbose: true, useDartAnalyzer: false);
```
evanweible-wf, line 223: should be able to remove `useDartAnalyzer: false` from these since that's the default value now

test/tools/analyze_tool_test.dart
```diff
@@ -163,19 +206,27 @@ void main() {
...
+    });
+
+    test('in useDartAnalyze mode', () {
+      expect(
+          Logger.root.onRecord, emitsThrough(infoLogOf(contains('dart -t a'))));
```
evanweible-wf, line 228: I think this should be:
```suggestion
          Logger.root.onRecord, emitsThrough(infoLogOf(contains('dart analyze -t a'))));
```

tool/dart_dev/config.dart
```diff
@@ -3,5 +3,8 @@ import 'package:glob/glob.dart';
 
 final config = {
   ...coreConfig,
-  'format': FormatTool()..exclude = [Glob('test/**/fixtures/**.dart')],
+  'analyze': AnalyzeTool()..useDartAnalyze = true,
+  'format': FormatTool()
+    ..exclude = [Glob('test/**/fixtures/**.dart')]
+    ..formatter = Formatter.dartFormat,
```
annawatson-wk, line 9: Do we want these for ourselves? I assumed so but I'm open to changing them back.

README.md
```diff
@@ -110,9 +110,10 @@ memorized or referenced in order to run said task.
 Consider formatting as an example. The default approach to formatting files is
 to run `dartfmt -w .`. But, some projects may want to exclude certain files that
 would otherwise be formatted by this command. Or, some projects may want to use
-`pub run dart_style:format` instead of `dartfmt`. Currently, there is no
-project-level configuration supported by the formatter, so these sorts of things
-just have to be documented in a `README.md` or `CONTRIBUTING.md`.
+`dart run dart_style:format` instead of `dartfmt` or `dart format` instead of 
+`dartfmt`. Currently, there is no project-level configuration supported by the
```
annawatson-wk, line 114: Does this line make sense?

evanweible-wf: Hmm.. since the purpose of this section is to explain why dart_dev itself is useful, I think we can leave it how it was except with `dartfmt` swapped for `dart format`, since the latter is now the canonical SDK tool.

README.md
```diff
@@ -110,9 +110,10 @@ memorized or referenced in order to run said task.
 Consider formatting as an example. The default approach to formatting files is
 to run `dartfmt -w .`. But, some projects may want to exclude certain files that
 would otherwise be formatted by this command. Or, some projects may want to use
-`pub run dart_style:format` instead of `dartfmt`. Currently, there is no
-project-level configuration supported by the formatter, so these sorts of things
-just have to be documented in a `README.md` or `CONTRIBUTING.md`.
+`dart run dart_style:format` instead of `dartfmt` or `dart format` instead of 
+`dartfmt`. Currently, there is no project-level configuration supported by the 
+formatter, so these sorts of things just have to be documented in a `README.md` 
+or `CONTRIBUTING.md`.
```
evanweible-wf, line 116:
```suggestion
`pub run dart_style:format` instead of `dart format`. Currently, there is no
project-level configuration supported by the formatter, so these sorts of things
just have to be documented in a `README.md` or `CONTRIBUTING.md`.
```

README.md
```diff
@@ -144,33 +145,33 @@ Using existing tooling provided by (or conventionalized by) the Dart community
...
-default of `pub run test`.
+will be run via `dart run build_runner test`, otherwise it falls back to the
+default of `dart run test`.
 
 ```bash
 # In a project without a `build_test` dependency:
 $ ddev test
 [INFO] Running subprocess:
-pub run test
+dart run test
```
evanweible-wf, line 166:
```suggestion
dart test
```

README.md
```diff
@@ -144,33 +145,33 @@ Using existing tooling provided by (or conventionalized by) the Dart community
...
-`pub run build_runner test`. Based on the project, you would need to know which
+`dart run build_runner test`. Based on the project, you would need to know which
 test command should be run.
 
 With `dart_dev`, the `TestTool` handles this automatically by checking the
 project's `pubspec.yaml` for a dependency on `build_test`. If present, tests
-will be run via `pub run build_runner test`, otherwise it falls back to the
-default of `pub run test`.
+will be run via `dart run build_runner test`, otherwise it falls back to the
+default of `dart run test`.
```
evanweible-wf, line 160:
```suggestion
default of `dart test`.
```

README.md
```diff
@@ -144,33 +145,33 @@ Using existing tooling provided by (or conventionalized by) the Dart community
 should always be the goal, but the reality is that there are gaps. Certain use
 cases can be made more convenient and new use cases may arise.
 
-Consider test running as an example. For simple projects, `pub run test` is
+Consider test running as an example. For simple projects, `dart run test` is
```
evanweible-wf, line 148:
```suggestion
Consider test running as an example. For simple projects, `dart test` is
```

README.md
```diff
@@ -144,33 +145,33 @@ Using existing tooling provided by (or conventionalized by) the Dart community
...
 sufficient. In fact, the test package supports a huge amount of project-level
 configuration via `dart_test.yaml`, which means that for projects that are
-properly configured, `pub run test` just works.
+properly configured, `dart run test` just works.
```
evanweible-wf, line 151:
```suggestion
properly configured, `dart test` just works.
```

lib/src/tools/test_tool.dart
```diff
@@ -19,9 +19,9 @@ final _log = Logger('Test');
 
 /// A dart_dev tool that runs dart tests for the current project.
 ///
-/// Tests will be run via `pub run test` unless the current project depends on
+/// Tests will be run via `dart run test` unless the current project depends on
```
evanweible-wf, line 22:
```suggestion
/// Tests will be run via `dart test` unless the current project depends on
```

PR #363, opened 2021-11-16
Compile the dart_dev run script for better performance

evanweible-wf:
Running a `dart_dev` command involves several subprocesses. The main reason for this is because we wanted to support rich package configuration via a `.dart` file that can do whatever is necessary and possible, rather than just a static config file in something like yaml. To make this work with `dart run dart_dev`, however, requires generating another Dart program that imports this config file from the current package and uses it when running `dart_dev`, which we then run via a subprocess.

The execution flow looks like this:

Screen Shot 2021-11-20 at 11 04 23 AM

With the exception that currently we don't perform the second precompilation step, meaning that every time you run `dart_dev`, you have to pay the cost of Dart compiling that generated run script. In some informal benchmarking, that step takes about **5 seconds**.

This PR introduces logic that will compile this run script and store the result in `.dart_tool/dart_dev/` so that subsequent runs can use it directly. Running precompiled Dart executables is very fast (~0.5s for kernel snapshots and ~0.02s for native executables), so once the compilation step is done, subsequent runs save almost all of the aforementioned 5 seconds.

The compiled result is cached along with a digest of its inputs. If any of these inputs change, the script will be recompiled:
- `.dart_tool/package_config.json` (pub get resolution + the current Dart SDK)
- `tool/dart_dev/config.dart`
- `tool/**.dart` (if the above config file has any relative imports)

Additionally, this compilation step will be skipped if:
- `tool/dart_dev/config.dart` imports from any package that is considered mutable

## Benchmarking

Here's some more informal benchmarking to demonstrate how this impacts first runs and subsequent (cached) runs:

### `dart run dart_dev` (first run)
| SDK                | Before | After  | Delta |
| ------------------ | ------ | ------ | ----- |
| 2.13.4             | ~15.6s | ~21s   | +5.4s |
| 2.14.4             | ~12.6s | ~18s   | +5.4s |
| 2.15.0-268.18.beta | ~11.7s | ~15.5s | +3.8s |

### `dart run dart_dev` (cached run)

| SDK                | Before | After | Delta |
| ------------------ | ------ | ----- | ----- |
| 2.13.4             | ~6.7s  | ~1.5s | -5.2s |
| 2.14.4             | ~7s    | ~2s   | -5s   |
| 2.15.0-268.18.beta | ~6.5s  | ~1.3s | -5.2s |

To summarize, the first run takes a perf hit of about 5 seconds, but as a result, subsequent runs have very little overhead and save that same 5 seconds every time.

---

This PR also adds a `clean` subcommand to `dart run dart_dev` that will remove the files in `.dart_tool/dart_dev/`, in case a bug is found in our compilation caching logic.

### DIFF
--- a/lib/src/dart_dev_runner.dart
+++ b/lib/src/dart_dev_runner.dart
@@ -5,6 +5,7 @@ import 'package:args/command_runner.dart';
 
 import 'dart_dev_tool.dart';
 import 'events.dart' as events;
+import 'tools/clean_tool.dart';
 import 'utils/version.dart';
 
 // import 'package:completion/completion.dart' as completion;
@@ -12,6 +13,10 @@ import 'utils/version.dart';
 class DartDevRunner extends CommandRunner<int> {
   DartDevRunner(Map<String, DevTool> commands)
       : super('dart_dev', 'Dart tool runner.') {
+    // For backwards-compatibility, only add the `clean` command if it doesn't
+    // conflict with any configured command.
+    commands.putIfAbsent('clean', () => CleanTool());
+
     commands.forEach((name, builder) {
       final command = builder.toCommand(name);
       if (command.name != name) {
@@ -19,6 +24,7 @@ class DartDevRunner extends CommandRunner<int> {
       }
       addCommand(command);
     });
+
     argParser
       ..addFlag('verbose',
           abbr: 'v', negatable: false, help: 'Enables verbose logging.')
--- a/lib/src/executable.dart
+++ b/lib/src/executable.dart
@@ -1,16 +1,21 @@
 import 'dart:async';
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:analyzer/dart/analysis/utilities.dart';
 import 'package:args/command_runner.dart';
+import 'package:crypto/crypto.dart';
 import 'package:dart_dev/dart_dev.dart';
 import 'package:dart_dev/src/dart_dev_tool.dart';
 import 'package:dart_dev/src/utils/format_tool_builder.dart';
 import 'package:dart_dev/src/utils/parse_flag_from_args.dart';
+import 'package:glob/glob.dart';
+import 'package:glob/list_local_fs.dart';
 import 'package:io/ansi.dart';
 import 'package:io/io.dart' show ExitCode;
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as p;
+import 'package:pubspec_parse/pubspec_parse.dart';
 
 import '../utils.dart';
 import 'dart_dev_runner.dart';
@@ -22,13 +27,18 @@ import 'utils/logging.dart';
 
 typedef _ConfigGetter = Map<String, DevTool> Function();
 
+const _packageConfigPath = '.dart_tool/package_config.json';
+final _packageConfig = File(_packageConfigPath);
+final _runExecutablePath = p.join(cacheDirPath, 'run');
+final _runExecutable = File(_runExecutablePath);
+final _runExecutableDigestPath = p.setExtension(_runExecutablePath, '.digest');
+final _runExecutableDigest = File(_runExecutableDigestPath);
 final _runScriptPath = p.join(cacheDirPath, 'run.dart');
-
 final _runScript = File(_runScriptPath);
-
 const _configPath = 'tool/dart_dev/config.dart';
-
+final _config = File(_configPath);
 const _oldDevDartPath = 'tool/dev.dart';
+final _oldDevDart = File(_oldDevDartPath);
 
 final _relativeDevDartPath = p.relative(
   p.absolute(_configPath),
@@ -37,29 +47,30 @@ final _relativeDevDartPath = p.relative(
 
 Future<void> run(List<String> args) async {
   attachLoggerToStdio(args);
-  final configExists = File(_configPath).existsSync();
-  final oldDevDartExists = File(_oldDevDartPath).existsSync();
 
-  if (!configExists) {
+  if (!_config.existsSync()) {
     log.fine('No custom `tool/dart_dev/config.dart` file found; '
         'using default config.');
   }
-  if (oldDevDartExists) {
+  if (_oldDevDart.existsSync()) {
     log.warning(yellow.wrap(
         'dart_dev v3 now expects configuration to be at `$_configPath`,\n'
         'but `$_oldDevDartPath` still exists. View the guide to see how to upgrade:\n'
         'https://github.com/Workiva/dart_dev/blob/master/doc/v3-upgrade-guide.md'));
   }
 
-  if (args.contains('hackFastFormat') && !oldDevDartExists) {
+  if (args.contains('hackFastFormat') && !_oldDevDart.existsSync()) {
     await handleFastFormat(args);
-
     return;
   }
 
-  generateRunScript();
+  final processArgs = generateRunScript();
   final process = await Process.start(
-      Platform.executable, [_runScriptPath, ...args],
+      processArgs.first,
+      [
+        if (processArgs.length > 1) ...processArgs.sublist(1),
+        ...args,
+      ],
       mode: ProcessStartMode.inheritStdio);
   ensureProcessExit(process);
   exitCode = await process.exitCode;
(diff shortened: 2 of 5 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/dart_dev_runner.dart
```diff
@@ -5,20 +5,26 @@ import 'package:args/command_runner.dart';
...
 import 'utils/version.dart';
 
 // import 'package:completion/completion.dart' as completion;
 
 class DartDevRunner extends CommandRunner<int> {
   DartDevRunner(Map<String, DevTool> commands)
       : super('dart_dev', 'Dart tool runner.') {
+    // For backwards-compatibility, only add the `clean` command if it doesn't
+    // conflict with any configured command.
+    commands.putIfAbsent('clean', () => CleanTool());
```
greglittlefield-wf, line 18: If this is a read-only map (`const` or otherwise), this will fail; can we make a copy of the map before adding this?

lib/src/executable.dart
```diff
@@ -37,29 +47,30 @@ final _relativeDevDartPath = p.relative(
...
     return;
   }
 
-  generateRunScript();
+  final processArgs = generateRunScript();
   final process = await Process.start(
-      Platform.executable, [_runScriptPath, ...args],
+      processArgs.first,
+      [
+        if (processArgs.length > 1) ...processArgs.sublist(1),
```
greglittlefield-wf, line 71: #nit This check is unnecessary, since for a length of 1 .sublist will return an empty list:
```suggestion
        ...processArgs.sublist(1),
```

lib/src/executable.dart
```diff
@@ -101,21 +112,103 @@ Future<void> handleFastFormat(List<String> args) async {
...
+  // Generate a digest of inputs to the run script. We use this to determine
+  // whether we need to recompile the executable.
+  String encodedDigest;
+  logTimedSync(log, 'Computing run script digest', () {
+    var configHasRelativeImports = false;
+    var configHasSamePackageImports = false;
+    if (_config.existsSync()) {
+      final contents = _config.readAsStringSync();
+      configHasRelativeImports =
+          RegExp(r'''^import ['"][^:]+''').hasMatch(contents);
```
greglittlefield-wf, line 136: I don't believe this regex is providing the test we want here.
1. `package:` imports will match this regex, with the match (`group(0)`) being `import 'package`
2. This only checks for imports on the first line of the file due to the use of `^` without the `multiLine: true`

lib/src/executable.dart
```diff
@@ -101,21 +112,103 @@ Future<void> handleFastFormat(List<String> args) async {
...
+      if (_runExecutable.existsSync()) {
+        _runExecutable.deleteSync();
+      }
+      if (_runExecutableDigest.existsSync()) {
+        _runExecutableDigest.deleteSync();
+      }
+      return;
+    }
+
+    final digest = md5.convert([
```
greglittlefield-wf, line 159: #nit it might be helpful to log which files are included in the digest, for debugging purposes.

Or at the very least, log the value of `configHasRelativeImports`.

lib/src/executable.dart
```diff
@@ -101,21 +112,103 @@ Future<void> handleFastFormat(List<String> args) async {
...
+      ..._packageConfig.readAsBytesSync(),
+      if (_config.existsSync()) ..._config.readAsBytesSync(),
+      if (configHasRelativeImports)
+        for (final file in Glob('tool/**.dart', recursive: true)
+            .listSync()
+            .whereType<File>()
+            .where((f) => f.path != _configPath))
```
greglittlefield-wf, line 166: Should we be canonicalizing these paths before comparing them? I'm not sure whether `Glob` makes any guarantees about the path formats in the returned files.

lib/src/executable.dart
```diff
@@ -101,21 +112,103 @@ Future<void> handleFastFormat(List<String> args) async {
...
+      // Delete the previous exectuable and digest so that if we hit a failure
+      // trying to compile, we don't leave the outdated one in place.
+      if (_runExecutable.existsSync()) {
+        _runExecutable.deleteSync();
+      }
+      if (_runExecutableDigest.existsSync()) {
+        _runExecutableDigest.deleteSync();
+      }
+
+      final args = ['compile', 'exe', _runScriptPath, '-o', _runExecutablePath];
```
greglittlefield-wf, line 186: It looks like Pub likely uses the `kernel` compile command; https://github.com/dart-lang/pub/blob/400f21e9883ce6555b66d3ef82f0b732ba9b9fc8/lib/src/dart.dart#L149-L150

I wonder why that's the one they chose...

Is there a reason you went with `exe` over any of other options? If so, could you add a comment?

lib/src/executable.dart
```diff
@@ -101,21 +112,103 @@ Future<void> handleFastFormat(List<String> args) async {
...
+        _runExecutable.deleteSync();
+      }
+      if (_runExecutableDigest.existsSync()) {
+        _runExecutableDigest.deleteSync();
+      }
+      return;
+    }
+
+    final digest = md5.convert([
+      ..._packageConfig.readAsBytesSync(),
```
greglittlefield-wf, line 160: I just thought about something... `path`-dependency packages can change without updating `package_config.dart`...

Depending on those within the `tool` directory should be fairly uncommon, but it could potentially cause issues if you're, say, developing a package that's being package-imported in the `tool/` directory, and are testing it out via a `path` dependency override...

I'm not sure how much we should worry about that case though, or what we'd be able to do about it without making this invalidation logic significantly more complex. 😕  What do you think?

greglittlefield-wf (changes requested):
Found a couple things, this looks awesome though!

lib/src/dart_dev_runner.dart
evanweible-wf: good catch!

lib/src/executable.dart
evanweible-wf: There wasn't a particular reason I chose `exe`, no, and now that I think about it, `kernel` is probably the better choice because we don't need this to be portable to systems that don't have the Dart SDK installed (we can run `dart .dart_tool/dart_dev/run.dill`). Maybe that will save some time on the compilation since (if I understand it correctly) it wouldn't need to include the SDK in the output?

evanweible-wf: Ok so I did some benchmarking:

### `dart compile`
How long it takes to compile the generated `run.dart` script. This is the "hit" users would take on first usage for the sake of faster subsequent, cached runs.

| SDK                | Command               | Time   |
| ------------------ | --------------------- | ------ |
| 2.13.4             | `dart compile exe`    | ~10.5s |
| 2.13.4             | `dart compile kernel` | ~9.8s  |
| 2.14.4             | `dart compile exe`    | ~11.5s |
| 2.14.4             | `dart compile kernel` | ~10.5s |
| 2.15.0-268.18.beta | `dart compile exe`    | ~11.5s |
| 2.15.0-268.18.beta | `dart compile kernel` | ~10.7s |

```ad-note
Note: dart_dev is not yet null-safe. Once migrated, we assume compilation time will improve, but probably not significantly.
```

### Running compiled Dart program
How long it takes to run the output of `dart compile`. In particular, we want to know if it's faster to run an `exe` or a `dill`. These times plus the compilation times will help determine which option is best.

| SDK                | Compiled Type                    | Time   |
| ------------------ | -------------------------------- | ------ |
| 2.13.4             | exe: `./dart_dev`                | ~0.02s |
| 2.13.4             | kernel: `dart dart_dev.dill`     | ~0.5s  |
| 2.13.4             | kernel: `dart run dart_dev.dill` | ~0.8s  |
| 2.14.4             | exe: `./dart_dev`                | ~0.02s |
| 2.14.4             | kernel: `dart dart_dev.dill`     | ~0.5s  |
| 2.14.4             | kernel: `dart run dart_dev.dill` | ~0.8s  |
| 2.15.0-268.18.beta | exe: `./dart_dev`                | ~0.02s |
| 2.15.0-268.18.beta | kernel: `dart dart_dev.dill`     | ~0.5s  |
| 2.15.0-268.18.beta | kernel: `dart run dart_dev.dill` | ~0.8s  |

```ad-note
Note: there is a perf hit (~0.9s) when running a Dart exe for the first time.
```

---

The tl;dr: here is that compiling a kernel dill is about 0.8-1s faster than a native executable, but running that dill takes about 0.5s every time compared to 0.02s for native executables (excluding first run). Based on that, I'm inclined to stick with `dart compile exe` because the first run is already going to be slow; might as well optimize for the best case to get the most speedup possible on every subsequent run.

lib/src/executable.dart
evanweible-wf: Good point.. I'm thinking it may be worth parsing the package_config.json and using it to lookup the type of dependency for each `package:` import in the dart_dev/config.dart file. If any of them are mutable, we skip the compilation & caching. This would also cover the same-package import scenario, which is nice because we're currently handling that separately.

greglittlefield-wf: Ooh good idea!

I think you might need the pubspec.lock to look up the dependency type, though, since package_config.json only contains paths to the packages being depended on and not the dependency type.

lib/src/dart_dev_runner.dart
alanknight-wk: Or you could write it as a literal with a spread (which I guess is another way of making a copy)
`  commands = {...commands, 'clean': () => CleanTool()};`

alexandercampbell-wk:
I've revived this PR here with merge conflicts resolved and most comments addressed: #426
Because of adjustments to our CI and the number of intervening changes to the tool, it seemed cleaner to open a fresh PR.
I'll close this PR for now— if it's preferable to move that commit into this PR and reopen, I'd be happy to make that change.

PR #364, opened 2021-12-09
CPLAT-16476 Allow args after separator to TestTool

evanweible-wf:
- Update `TestTool` to allow arguments after a separator (`--`). These arguments will always be passed to the `dart test` process. The main use case for this is integration with IDE plugins that enable running tests directly from the IDE.
- Update `FunctionTool` to allow arguments after a separator (`--`). There isn't a strong reason to disallow this since the function tool could do anything it wants with those args (and now we have a concrete use case for just that).
- Fix a bug in `takeAllArgs` (the arg mapper util used with `CompoundTool`) so that it now properly restores the first separator (`--`) if present in the original arguments list.

### DIFF
--- a/lib/src/utils/rest_args_with_separator.dart
+++ b/lib/src/utils/rest_args_with_separator.dart
@@ -25,20 +25,20 @@ List<String> restArgsWithSeparator(ArgResults argResults) {
 
   final args = argResults.arguments;
   final rest = argResults.rest;
-  var rCursor = 0;
-  for (var aCursor = 0; aCursor < args.length; aCursor++) {
+  var restIndex = 0;
+  for (var argsIndex = 0; argsIndex < args.length; argsIndex++) {
     // Iterate through the original args until we hit the first separator.
-    if (args[aCursor] == '--') break;
+    if (args[argsIndex] == '--') break;
     // While doing so, move a cursor through the rest args list each time we
     // match up between the original list and the rest args list. This works
     // because the rest args list should be an ordered subset of the original
     // args list.
-    if (args[aCursor] == rest[rCursor]) {
-      rCursor++;
+    if (args[argsIndex] == rest[restIndex]) {
+      restIndex++;
     }
   }
 
-  // At this point, [rCursor] should be pointing to the spot where the first arg
-  // separator should be restored.
-  return [...rest.sublist(0, rCursor), '--', ...rest.sublist(rCursor)];
+  // At this point, [restIndex] should be pointing to the spot where the first
+  // arg separator should be restored.
+  return [...rest.sublist(0, restIndex), '--', ...rest.sublist(restIndex)];
 }
--- a/lib/src/utils/rest_args_with_separator.dart
+++ b/lib/src/utils/rest_args_with_separator.dart
@@ -1,0 +1,44 @@
+import 'package:args/args.dart';
+
+/// Returns the "rest" args from [argResults], but with the arg separator "--"
+/// restored to its original position if it was included.
+///
+/// This is necessary because [ArgResults.rest] will _not_ include the separator
+/// unless it stopped parsing before it reached the separator.
+///
+/// The use case for this is a [CompoundTool] that uses the [takeAllArgs] arg
+/// mapper, because the goal there is to forward on the original args minus the
+/// consumed options and flags. If the separator has also been removed, you may
+/// hit an error when trying to parse those args.
+///
+///     var parser = ArgParser()..addFlag('verbose', abbr: 'v');
+///     var results = parser.parse(['a', '-v', 'b', '--', '--unknown', 'c']);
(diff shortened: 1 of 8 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

todbachman-wf (dismissed):

todbachman-wf (approved):

lib/src/utils/rest_args_with_separator.dart
```diff
@@ -0,0 +1,44 @@
...
+/// mapper, because the goal there is to forward on the original args minus the
+/// consumed options and flags. If the separator has also been removed, you may
+/// hit an error when trying to parse those args.
+///
+///     var parser = ArgParser()..addFlag('verbose', abbr: 'v');
+///     var results = parser.parse(['a', '-v', 'b', '--', '--unknown', 'c']);
+///     print(results.rest);
+///     // ['a', 'b', '--unknown', 'c']
+///     print(restArgsWithSeparator(results));
+///     // ['a', 'b', '--', '--unknown', '-c']
```
alanknight-wk, line 19: Is this correct? It turns from `c` in `results` into `-c` here?

lib/src/utils/rest_args_with_separator.dart
```diff
@@ -0,0 +1,44 @@
...
+List<String> restArgsWithSeparator(ArgResults argResults) {
+  // If no separator was used, return the rest args as is.
+  if (!argResults.arguments.contains('--')) {
+    return argResults.rest;
+  }
+
+  final args = argResults.arguments;
+  final rest = argResults.rest;
+  var rCursor = 0;
```
alanknight-wk, line 28: Maybe a clearer name? e.g. argSeparatorPosition

lib/src/utils/rest_args_with_separator.dart
```diff
@@ -0,0 +1,44 @@
...
+  for (var aCursor = 0; aCursor < args.length; aCursor++) {
+    // Iterate through the original args until we hit the first separator.
```
alanknight-wk, line 30: I'm having trouble following this. It seems like the only result of this is setting rCursor. But isn't rCursor just `args.indexOf('--')`?

alanknight-wk (approved):

lib/src/utils/rest_args_with_separator.dart
evanweible-wf: Oh shoot, that's incorrect. Should just be `c` still - I'll fix that. Good catch!

lib/src/utils/rest_args_with_separator.dart
evanweible-wf: That would be the position of the separator within the original arguments list, which isn't necessarily the same as the position at which the separator should be restored in the `rest` arguments list, since the former excludes flags and options that were parsed by the ArgParser. So for example:

```dart
final argParser = ArgParser()
  ..addOption('output')
  ..addFlag('verbose');
final argResults = argParser.parse(['foo', '--output', 'out', '--verbose', '--', 'bar', '--baz']);
print(argResults.arguments);
// ['foo', '--output', 'out', '--verbose', '--', 'bar', '--baz']
// In this list, separator index is 4
print(argResults.rest);
// ['foo', 'bar', '--baz']
// In this list, we want to restore the separator at index 1
```

lib/src/utils/rest_args_with_separator.dart
evanweible-wf: I'm a little torn on this. I originally had something like what you suggested, but then usages like this feel a bit more confusing:

```
if (args[aCursor] == rest[separatorArgPosition]) {
```

lib/src/utils/rest_args_with_separator.dart
alanknight-wk: OK, I get it. Ugh. It'd be better if ArgParser just didn't think it knew about '--' and included it.

lib/src/utils/rest_args_with_separator.dart
alanknight-wk: Yeah, I see that. What about not abbreviating. argsCursor and restCursor. I see aCursor at a glance and read it as "an instance of Cursor". Or argsIndex and restIndex?

alanknight-wk (approved):

alanknight-wk (approved):

alanknight-wk (dismissed):

evanweible-wf:
QA +1

- [x] Tested this locally in one of our projects that uses dart_dev + test_html_builder with browser tests and was able to run and debug tests via the links in the editor.

todbachman-wf (approved):

evanweible-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #369, opened 2022-01-27
Null safety

robbecker-wf:
# Null safety migration DRAFT

I ran the migration tool and took the suggested changes, then fixed up about 5 errors and a few unused imports.
This will need a solid pass to make sure we didn't break any public API, fix any weird things that the migration tool did that we want to improve, and test it


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/executable.dart
```diff
@@ -130,8 +130,7 @@ void main(List<String> args) async {
 ''';
 }
 
-Future<void> runWithConfig(
-    List<String> args, _ConfigGetter configGetter) async {
+Future<int> runWithConfig(List<String> args, _ConfigGetter configGetter) async {
```
robbecker-wf, line 133: this was returning the exitcode before, but only if there was a handled exception, so I changed the return type to in and added a final return 0 for success.

PR #373, opened 2022-05-11
Fix running dart_dev in mixed-version packages opted into null safety

greglittlefield-wf:
## Motivation
dart_dev currently fails to run in mixed-version packages that have opted into null safety but have some non-null-safe dependencies.

This is because, in those packages, the bootstrap script is run in sound null safety mode by default, which fails because dart_dev (and potentially other packages pulled in via the config) are not themselves null-safe.

This is what those errors look like:
```
$ ddev format
Error: Cannot run with sound null safety, because the following dependencies
don't support null safety:

 - package:dart_dev

For solutions, see https://dart.dev/go/unsound-null-safety
```

When using a config.dart that's also opted out of null-safety via a language comment:
```
$ ddev format
tool/dart_dev/config.dart:1:1: Error: A library can't opt out of null safety by default, when using sound null safety.
// @dart=2.9
^^^^^^^^^^^^
Error: Cannot run with sound null safety, because the following dependencies
don't support null safety:

 - package:dart_dev

For solutions, see https://dart.dev/go/unsound-null-safety
```

## Solution
- While dart_dev is null safe, opt the run script out of null safety using a version comment (this could also be achieved via a `--no-sound-null-safety` flag, but this seemed more straightforward)
- If there's a language version comment in `config.dart`, copy that over to the run script. This will ensure that non-null-safe configs will continue to work even after dart_dev itself is migrated to null safety,
- Add functional tests that fail unless this new logic is in place

## Testing
- CI passes

### DIFF
--- a/.github/workflows/dart_ci.yml
+++ b/.github/workflows/dart_ci.yml
@@ -7,7 +7,7 @@ on:
       - 'test_consume_*'
   pull_request:
     branches:
-      - '*'
+      - '**'
 
 jobs:
   build:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 - Fix errors when running in mixed-version packages that have opted into null safety via pubspec.yaml.
 
+## [3.8.4](https://github.com/Workiva/dart_dev/compare/3.8.3...3.8.4)
+- Replace deprecated CLI command.
+
+## [3.8.3](https://github.com/Workiva/dart_dev/compare/3.8.2...3.8.3)
+- CI improvement.
+
 ## [3.8.2](https://github.com/Workiva/dart_dev/compare/3.8.1...3.8.2)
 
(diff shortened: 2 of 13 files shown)


aviary-wf:
## Security Insights

## Action Items
---

evanweible-wf (approved):

evanweible-wf:
LGTM - thanks for adding this @greglittlefield-wf! Would you mind adding a changelog entry for this? I think it makes sense as a patch, so the next version would be 3.8.4

greglittlefield-wf:
@evanweible-wf Np, thanks for reviewing! Can do!

evanweible-wf (approved):

evanweible-wf:
QA +1
- [x] CI passes

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #381, opened 2022-07-14
Add import sorting to format tool (backpatch to v3.7.1)

regenvanwalbeek-wf:
Note: This is a backpatch to 3.7.1 for consumers who have not upgraded to analyzer 1.0.0 - The changes for the mainline branch exist [here](https://github.com/Workiva/dart_dev/pull/382)

Having to manually sort imports is tedious. It would be better if we could just run the formatter to receive sorted imports.

I added an `organizeImports` flag to the `FormatTool`. When set to true, it will also run the file through an import sorter.

The import sorter tries to comply with https://dart.dev/tools/linter-rules#directives_ordering


lib/src/executable.dart
```diff
@@ -58,6 +58,11 @@ Future<void> run(List<String> args) async {
     return;
   }
 
+  if (args.contains('sort_imports')) {
+    await handleSortImports(args);
+    return;
+  }
```
regenvanwalbeek-wf, line 64: Open to suggestions on how the FormatTool should kick off the import sorter.

`FormatTool#buildExecution` returns a `FormatExecution`, which is a declarative representation of `FormatTool` so that `FormatTool` can be easily tested. For this reason, I tried to keep import sorting out of the format tool itself, and chose to allow `dart_dev` to handle import sorting.

pubspec.yaml
```diff
@@ -24,8 +22,15 @@ dependencies:
...
+  meta: ">=1.2.2 <1.7.0" # Workaround to avoid https://github.com/dart-lang/sdk/issues/46142
   test: ^1.15.7
   test_descriptor: ^1.2.0
   test_process: ^1.0.4
+
+dependency_validator:
+ exclude:
+  - test/tools/fixtures/sort_imports/imports.dart
+ ignore:
+  - meta
```
regenvanwalbeek-wf, line 36: Looking for guidance here. During the `format_and_publish` check:

```
Package validation found the following potential issue:
* "dependency_validator" is not a key recognized by pub - did you mean "dependency_overrides"?
```

Looks like we didn't add support for `dart_dependency_validator.yaml` until `dependency_validator` 3.0.0

.github/workflows/dart_ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [ 2.7.2, 2.13.4, stable, dev ]
+        sdk: [ 2.7.2, 2.13.4 ]
```
regenvanwalbeek-wf, line 18: Looking for sanity check - I am making an assumption that the backpatch release line will not be usable on stable/dev.

regenvanwalbeek-wf:
Will I have to add a `skynet.yaml` for this to successfully release?

.github/workflows/dart_ci.yml
evanweible-wf: Yeah, this is reasonable 👍

lib/src/executable.dart
evanweible-wf: I think my preference would be to handle this within the `FormatTool` rather than adding a hidden `sort_imports` command. I know we had to do that to support formatting from an IDE, but in this case I think we can accomplish it without.

I'm thinking we could keep FormatExecution the way it is in master so that it still just represents the `dart format`/`dartfmt` process, but then conditionally spawn the `ImportCleanerTool` within the `run()`.

```dart
  @override
  FutureOr<int> run([DevToolExecutionContext context]) async {
    context ??= DevToolExecutionContext();
    final formatExecution = buildExecution(
      context,
      configuredFormatterArgs: formatterArgs,
      defaultMode: defaultMode,
      exclude: exclude,
      formatter: formatter,
    );
    if (formatExecution.exitCode != null) {
      return formatExecution.exitCode;
    }
    var exitCode = await runProcessAndEnsureExit(formatExecution.process, log: _log);
    if (exitCode != 0) {
      return exitCode;
    }
    if (organizeImports) {
      exitCode = await ImportCleanerTool().run(context);
    }
    return exitCode;
  }
```

That way, we're only really adding one conditional branch of logic here and we can still unit test the `ImportCleanerTool` separately. If we want, we could add a couple of functional tests for formatting - one with and one without `organizeImports`.

lib/src/tools/format_tool.dart
```diff
@@ -68,6 +68,11 @@ class FormatTool extends DevTool {
   /// Run `dartfmt -h -v` to see all available args.
   List<String> formatterArgs;
 
+  /// If the formatter should also organize imports.
+  ///
+  /// By default, this is disabled.
+  bool organizeImports = false;
```
evanweible-wf, line 74: I like the name `organizeImports` 👍 It would be nice to name everything consistently, though. Could we rename `ImportCleanerTool` to `ImportOrganizerTool`?

lib/src/executable.dart
evanweible-wf: I guess we wouldn't be able to pass the args for the format tool directly through to the `ImportCleanerTool`. We could update the `context` that we pass through like so:

```dart
    if (organizeImports) {
      final importTool = ImportCleanerTool();
      context = context.update(argResults: importTool.argParser.parse([ ... ]);
      exitCode = await importTool.run(context);
    }
```

But that would require some refactoring to avoid duplicating the work to parse the formatter tool's args. Since we're not actually making `ImportCleanerTool` public and we're exposing this behavior via a flag on the `FormatterTool`, maybe we don't need the tool wrapper for this logic at all. We could instead just have this tool call into a utility that does the import organizing directly rather than constructing a tool and running it ourselves.

Maybe something like this? https://github.com/Workiva/dart_dev/compare/import_sort...import_sort_no_cleaner_tool?expand=1

lib/src/utils/import_cleaner/import_cleaner.dart
```diff
@@ -0,0 +1,148 @@
+import 'package:analyzer/dart/analysis/utilities.dart';
+import 'package:analyzer/dart/ast/token.dart';
+
+import 'import.dart';
+import 'import_collector.dart';
+
+/// Takes in a file as a string and cleans up the imports. Sorts imports and removes double quotes.
+///
+/// Throws an ArgumentError if [sourceFileContents] cannot be parsed.
+String cleanImports(String sourceFileContents) {
```
evanweible-wf, line 10: Same note in this file - would be nice to use a consistent name for this.

pubspec.yaml
evanweible-wf: Yeah I think for this backpatch we should just disable the pub publish CI check.

pubspec.yaml
```diff
@@ -11,10 +11,8 @@ dependencies:
   args: ^1.5.2
   async: ^2.3.0
   build_runner: ^1.7.0
-  completion: ^0.2.1+1
   glob: ^1.1.7
   io: ^0.3.3
-  meta: ">=1.2.2 <1.7.0" # Workaround to avoid https://github.com/dart-lang/sdk/issues/46142
```
evanweible-wf, line 17: I think we may have made this a regular dependency here to avoid needing the same workaround everywhere else since `dart_dev` was such a common dependency. Would you mind keeping this just in case others happen to be relying on it transitively?

regenvanwalbeek-wf: That makes sense, will revert.

lib/src/tools/format_tool.dart
```diff
@@ -283,25 +306,41 @@ class FormatterInputs {
...
+      : exitCode = null;
 
   /// If non-null, the execution is already complete and the [FormatTool] should
   /// exit with this code.
   ///
   /// If null, there is more work to do.
   final int exitCode;
 
-  /// A declarative representation of the formatter process that should be run.
+  /// A declarative representation of the formatter processes that should be run.
```
evanweible-wf, line 327: #nit 
```suggestion
  /// A declarative representation of the formatter process that should be run.
```

lib/src/tools/format_tool.dart
```diff
@@ -283,25 +306,41 @@ class FormatterInputs {
...
   ///
-  /// This process' result should become the final result of the [FormatTool].
-  final ProcessDeclaration process;
+  /// If all processes result in exit code 0, [FormatTool] will return with exit code 0.
+  /// Otherwise, the first non-zero exit code will become the final result of the [FormatTool].
```
evanweible-wf, line 330:
```suggestion
  /// If this process results in a non-zero exit code, [FormatTool] should return it.
```

lib/src/utils/organize_imports/import_collector.dart
```diff
@@ -0,0 +1,23 @@
...
+    _imports.add(Import(node));
+    return null;
+  }
+
+  @override
+  List<Import> visitCompilationUnit(CompilationUnit node) {
+    node.visitChildren(this);
+    return _imports;
+  }
+}
```
evanweible-wf, line 23: Any reason not to collect and organize export directives, as well? We'd need to bucket and organize them separately, but otherwise I'd think the logic would be the same.

lib/src/utils/organize_imports/organize_imports_in_paths.dart
```diff
@@ -0,0 +1,126 @@
...
+  bool check = false,
+}) {
+  for (final path in paths) {
+    final exitCode =
+        _organizeImportsInFile(path, verbose: verbose, check: check);
+    if (exitCode != 0) {
+      return exitCode;
+    }
+  }
+  return 0;
```
evanweible-wf, line 60: #nit not a big deal, but we could avoid returning early here so that everything that is valid still gets organized and the user receives an error message for each file that failed - this is actually what `dart format` does when there are syntax errors.

```suggestion
  var exitCode = 0;
  for (final path in paths) {
    final codeForFile =
        _organizeImportsInFile(path, verbose: verbose, check: check);
    if (codeForFile != 0) {
      exitCode = codeForFile;
    }
  }
  return exitCode;
```

lib/src/utils/organize_imports/import_collector.dart
regenvanwalbeek-wf: oooh I can look at adding that, I don't think it will be too bad. We just hadn't had as many `#nit export order` code review comments, so it never crossed my mind

lib/src/tools/format_tool.dart
regenvanwalbeek-wf: @evanweible-wf now that this organizes imports and exports, should I just call this `organizeDirectives`?

evanweible-wf: Ah yeah, good call - I like `organizeDirectives` 👍

evanweible-wf (approved):

evanweible-wf:
@regenvanwalbeek-wf could you add a changelog entry for 3.7.2? I can't quite remember how backpatch releases work, so it's possible this may end up effectively being the release PR. I've updated MARV with the correct version.

regenvanwalbeek-wf:
🔍 Should this be a `3.8.0` since I'm adding a field to the public API? Probably not a huge deal

evanweible-wf:
Unfortunately since we're back patching, we're a bit limited since 3.8.x already exists. Ideally we would only be backpatching bugs, but I think we're alright to just call this a patch release anyway.

regenvanwalbeek-wf:
That makes sense. Pushed!

evanweible-wf (approved):

evanweible-wf:
QA +1
- [x] CI passes (tests added and updated in this PR)
- [x] Tested this locally with `organizeDirectives = false` and `organizeDirectives = true`

evanweible-wf:
@Workiva/release-management-p

todbachman-wf (approved):

rmconsole-wf (approved):
+1 from RM

PR #395, opened 2023-01-26
Change defaults

robbecker-wf:
Change the default formatter and analyzer to the newer options `dart format` and `dart analyze`

### DIFF
--- a/lib/src/tools/analyze_tool.dart
+++ b/lib/src/tools/analyze_tool.dart
@@ -74,7 +74,7 @@ class AnalyzeTool extends DevTool {
 
   @override
   FutureOr<int> run([DevToolExecutionContext context]) {
-    useDartAnalyze ??= false;
+    useDartAnalyze ??= true;
     return runProcessAndEnsureExit(
         buildProcess(context ?? DevToolExecutionContext(),
             configuredAnalyzerArgs: analyzerArgs,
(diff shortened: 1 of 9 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

evanweible-wf (dismissed):

dustinlessard-wf (dismissed):

robbecker-wf:
QA+1 CI Passes

lib/src/tools/analyze_tool.dart
```diff
@@ -74,7 +74,7 @@ class AnalyzeTool extends DevTool {
 
   @override
   FutureOr<int> run([DevToolExecutionContext context]) {
-    useDartAnalyze ??= false;
+    useDartAnalyze ??= true;
```
todbachman-wf, line 77: Maybe we should update the doc comment on `useDartAnalyze` to indicate that we will actually use `dart analyze` by default.

robbecker-wf: Oh that's a good idea!

todbachman-wf (approved):

dustinlessard-wf (changes requested):
Looks like the tests are not expecting 'analyze' to be in the buildArgs.

robbecker-wf:
@dustinlessard-wf Indeed. It looks like the analyze args are different depending on if you are using "dart analyze" mode vs "dartanalyzer" mode. Instead of fixing the tests, I think I'll investigate changing the code so that the analyzer args are the same no matter which mode you are in.

PR #406, opened 2023-06-08
Make run return nonnullable FutureOr<int>

robbecker-wf:
When migrating to null safety we missed making the run method return a non nullable FutureOr<int> ... and that is causing weird casts and slight type mismatches in other tools.  This fixes that since the exit code is either zero pass and non-zero for fail but should never be null.

### DIFF
--- a/.github/workflows/dart_ci.yml
+++ b/.github/workflows/dart_ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [ 2.18.7, stable ]
+        sdk: [ 2.18.7, 2.19.6 ]
     steps:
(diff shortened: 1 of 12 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

.github/workflows/dart_ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: dart-lang/setup-dart@v1
         with:
-          sdk: stable
+          sdk: 2.18.7
```
robbecker-wf, line 47: since 3.0 got released to stable, it now doesn't work .. so avoiding that for now.

PR #420, opened 2023-11-08
FEDX-518 Now that dart_dev itself is null-safe, it should be able to generate a null-safe run script

evanweible-wf:
## Problem

When we migrated this package to null-safe, we missed some logic in our run script generation that would include a Dart language comment to set the language version to 2.9 by default. This prevents the null-safe dart_dev from being able to run on Dart 3 where null-safety is required.

## Solution

Remove this logic as it's no longer needed. We'll still copy over a language comment from `tool/dart_dev/config.dart` if one exists, otherwise we omit it which will allow dart_dev to run on Dart 3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 4.0.3
+
+- Fix generation of the run script that the `dart_dev` CLI uses so that it can
+run with sound null safety (and thus run on Dart 3).
+
 ## 4.0.1
 
 - Fix type mismatch between the expected return type of the function passed to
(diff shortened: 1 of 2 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

greglittlefield-wf (approved):
- Changes look good
- Existing functional  [tests](https://github.com/Workiva/dart_dev/blob/9e6f3e87087c3e1a7d06560c6331acdd21f5cee4/test/functional/null_safety_functional_test.dart#L8) pass, that verify that dart_dev works as expected when packages have opted into null safety
- Manually tested in a package that is not yet opted into null safety and has a custom config, and dart_dev commands work as expected

+10

evanweible-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #424, opened 2024-02-20
format_tool: optimization pass

alexandercampbell-wk:
We can save a lot (~90%) of execution time of `dart_dev format` on some repos with two changes:

 - Not recursing into directories that are a priori excluded from the formatter's list (e.g., hidden directories like .git). Note that this means we no longer build and log a comprehensive list of excluded files. I think it's worth it to run so much faster.

 - Removing the `collapseDirectories` feature, to be replaced with batch parallel invocations. Computing the collapsible directories was some kind of superlinear deduplication. This could maybe be optimized in a way that would allow it to remain, but it seems it was intended as a stopgap solution for argv limitations in the first place.

This is technically a breaking change to the API since we no longer return a full list of excluded files and hidden directories.
I have retained the `collapseDirectories` parameter in the name of backward compatibility for now, but it's ignored and marked `@deprecated`.

This moves the responsibility for handling argv limitations into `FormatTool`.

### DIFF
--- a/lib/src/tools/format_tool.dart
+++ b/lib/src/tools/format_tool.dart
@@ -3,7 +3,6 @@ import 'dart:io';
 
 import 'package:args/args.dart';
 import 'package:glob/glob.dart';
-import 'package:glob/list_local_fs.dart';
 import 'package:io/ansi.dart';
 import 'package:io/io.dart' show ExitCode;
 import 'package:logging/logging.dart';
@@ -140,6 +139,23 @@ class FormatTool extends DevTool {
     return exitCode;
   }
 
+  // Similar to listSync() but does not recurse into hidden directories.
+  static List<FileSystemEntity> _listSyncWithoutHidden(Directory dir,
+      {required bool recursive, required bool followLinks}) {
+    var allEntries = <FileSystemEntity>[];
+    dir.listSync(recursive: false, followLinks: followLinks).forEach((element) {
+      final basename = p.basename(element.path);
+      if (basename.length > 1 && basename.startsWith(".")) return;
+
+      allEntries.add(element);
+      if (element is Directory) {
+        allEntries.addAll(_listSyncWithoutHidden(element,
+            recursive: recursive, followLinks: followLinks));
+      }
+    });
+    return allEntries;
+  }
+
   /// Builds and returns the object that contains:
   /// - The file paths
   /// - The paths that were excluded by an exclude glob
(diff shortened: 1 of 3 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

evanweible-wf (commented):
LGTM! Feel free to add the filter functions as alternatives to globs in this PR or a separate one.

lib/src/tools/format_tool.dart
```diff
@@ -178,135 +192,37 @@ class FormatTool extends DevTool {
...
-        }
-
-        if (entry is File && !skipFilesInDirectory) {
-          _log.finest("adding $relative\n");
-          includedFiles.add(relative);
-        }
+      if (exclude.any((glob) => glob.matches(filename))) {
+        _log.finer('excluding $filename\n');
+      } else if (entry is File) {
+        _log.finest("adding $filename\n");
```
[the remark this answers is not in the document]
alexandercampbell-wk: Good call. done.

alexandercampbell-wk:
Oops, I broke CI. I'll get that addressed today.

lib/src/tools/format_tool.dart
```diff
@@ -178,135 +191,51 @@ class FormatTool extends DevTool {
...
-      }
+    for (final entry in _listSyncWithoutHidden(dir,
+        recursive: true, followLinks: followLinks)) {
+      final filename = p.relative(entry.path, from: dir.path);
+      _log.finest('== Processing relative $filename ==\n');
 
       if (entry is Link) {
-        _log.finer('skipping link $relative\n');
-        skippedLinks.add(relative);
+        _log.finer('skipping link $filename\n');
```
tomconnell-wf, line 200: Why did these checks not make it into the listSync forEach?  Would they even be faster there?

tomconnell-wf (approved):

lib/src/tools/format_tool.dart
alexandercampbell-wk: That would definitely be faster— I chose to keep this closer to the original logic for now to avoid churn. I tentatively plan to make another pass to reduce the amount of work being done in this function later.

alexandercampbell-wk:

rmconsole-wf (approved):
+1 from RM

tomconnell-wf:
Rosie doesn't require QA on this repo?  :sus:

PR #442, opened 2024-11-03
Fix webdev version check for Dart 3

robbecker-wf:
# Summary

When running dart run dart_dev serve under Dart 3, it checks that webdev is installed, but it insists on webdev ^2.0.0. Dor Dart 3, only webdev ^3.0.0 resolves. 

# Changes
Switch the version to check for based on the version of Dart being run. For Dart 2, use ^2.0.0 and Dart 3 use ^3.0.0 of webdev.

### DIFF
--- a/lib/src/tools/webdev_serve_tool.dart
+++ b/lib/src/tools/webdev_serve_tool.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:args/args.dart';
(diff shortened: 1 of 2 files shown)


dustinlessard-wf (approved):

robbecker-wf:
QA+1 CI passes. Tested with Dart 2.19 and 3.5.4 locally. @Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

